PR #73, opened 2026-03-27
feat(flux2): GGUF transformer support, BF16 encoder fix, production quality

jamesbrink:
## Summary

Brings Flux.2 Klein from alpha to production quality by fixing three critical issues:

- **Native BF16 Qwen3 encoder** with proper multi-layer extraction (layers 9/18/27) — the upstream `ZImageTextEncoder` only returned the penultimate layer, so BF16 users got three identical copies instead of distinct layer representations
- **GGUF quantized transformer loading** (Q8/Q6/Q4) — was completely missing, causing "header too large" errors. Dequantizes to BF16 at load time using `candle_nn::Linear` for identical inference precision
- **Fixed scale/shift swap in GGUF final layer** — BFL format outputs `(shift, scale)` but code was treating it as diffusers' `(scale, shift)`, causing 3x output amplitude divergence and visible dithering artifacts

Also removes `[alpha]` status, updates docs/skills, and adds guidance logging.

## Test plan

- [x] `cargo fmt --check` passes
- [x] `cargo clippy -- -D warnings` passes  
- [x] `cargo test --workspace` — 644 tests pass (including new `qlast_layer_uses_bfl_shift_scale_ordering` test)
- [x] `mold run flux2-klein:q8 "prompt" --preview` produces photorealistic output
- [x] `mold run flux2-klein:bf16 "prompt" --preview` produces photorealistic output
- [x] GGUF VRAM usage matches BF16 (dequantizes to gpu_dtype, not F32)
- [x] Bare `flux2-klein` resolves to `:q8` (preserves existing installs)

### DIFF
--- a/.claude/skills/mold/SKILL.md
+++ b/.claude/skills/mold/SKILL.md
@@ -100,13 +100,15 @@ Default model if none specified: `flux-schnell:q8`
 
 **Z-Image**: `z-image-turbo:bf16`, `z-image-turbo:q8`, `z-image-turbo:q6`, `z-image-turbo:q4`
 
-**Alpha**: `flux2-klein:bf16`, `flux2-klein:q8`, `flux2-klein:q6`, `flux2-klein:q4`, `qwen-image:bf16`, `qwen-image:q8`, `qwen-image:q6`, `qwen-image:q4`, `wuerstchen-v2:fp16`
+**Flux.2 Klein**: `flux2-klein:bf16`, `flux2-klein:q8`, `flux2-klein:q6`, `flux2-klein:q4`
+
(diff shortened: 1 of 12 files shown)


jamesbrink:
All three Greptile findings addressed in bdcb919:

**P2 — Test doesn't assert BFL ordering**: Rewrote `qlast_layer_uses_bfl_shift_scale_ordering` to use non-uniform input where BFL `(shift, scale)` and diffusers `(scale, shift)` produce numerically different results, and assert the output is consistent with additive shift (BFL ordering).

**P2 — Duplicate transformer loading block**: Extracted `load_transformer()` helper method that handles GGUF detection and dispatch. Both eager and sequential paths now call this single method.

**P2 — Silent under-collection with out-of-bounds layer_indices**: Added explicit early validation in both `Bf16Qwen3Encoder::forward_with_layers` and `GgufQwen3Encoder::forward_with_layers` — bail on empty slice or out-of-bounds indices.

PR #112, opened 2026-03-31
fix(download): skip already-downloaded files on pull retry + fix qwen-image manifest

jamesbrink:
## Summary

- **Skip already-downloaded files on retry**: When pulling a model with multiple files, if one file fails (e.g. 404), previously downloaded files are now recognized on retry via size + SHA-256 verification, so only the failed files are re-downloaded.
- **Fix qwen-image:bf16 manifest**: The transformer shard filenames were wrong (2 shards: `00001-of-00002`, `00002-of-00002`) — the actual `Qwen/Qwen-Image-2512` repo has 9 shards (`00001-of-00009` through `00009-of-00009`). This caused a 404 on every pull attempt.

## Test plan

- [x] `cargo check` passes (full workspace)
- [x] `cargo test -p mold-ai-core` — all 364 tests pass
- [x] `cargo clippy -p mold-ai-core -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [ ] `mold pull qwen-image:bf16` downloads all 9 transformer shards successfully
- [ ] Interrupt a pull mid-download, retry — already-downloaded files are skipped

### DIFF
--- a/crates/mold-core/src/download.rs
+++ b/crates/mold-core/src/download.rs
@@ -420,9 +420,11 @@ impl Progress for CallbackProgress {
     }
 }
 
-/// Check if a file is already placed at its clean path with the expected size
-/// and passes integrity verification (if a SHA-256 hash is available).
-/// Used to skip re-downloading files that succeeded in a previous (partial) pull.
+/// Returns `true` if the file already exists at `clean_path` with the correct
+/// size and (if a SHA-256 is available) the correct digest.
+///
+/// **Side-effect**: if the file exists with matching size but failing integrity,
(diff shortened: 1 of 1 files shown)


jamesbrink:
Addressed Greptile findings:

| 1 | P2 | Destructive side-effect inside is_already_placed predicate | Fixed -- added doc-comment documenting the file-deletion side-effect |
| 2 | P2 | Skipped files produce non-sequential file_index in callbacks | Fixed -- pre-compute files needing download, use sequential download_idx counter |

PR #207, opened 2026-04-07
fix(qwen-image): improve Metal GGUF denoising performance

### LINKED ISSUE #202: Qwen-Image denoising extremely slow on Apple Metal (M4 Max, 0.04 it/s)
## Description

Qwen-Image denoising is extremely slow on Apple Silicon Metal — approximately **0.04–0.05 it/s** during the denoising loop. At 50 steps this means **~16+ minutes per image**, which is unusable.

## Hardware

- **Chip**: Apple M4 Max (16-core CPU, 40-core GPU)
- **Memory**: 48 GB unified
- **OS**: macOS (Darwin 25.3.0)

## Observed Performance

```
⠠ Denoising [━━━━━╸────────────────────────] 9/50  [00:03:18, 0.04 it/s]
⠐ Denoising [━━━━━━╸───────────────────────] 10/50 [00:03:21, 0.05 it/s]
```

At this rate, a full 50-step generation takes ~16–17 minutes. For reference, FLUX and other pipelines on the same hardware are significantly faster.

## Expected Performance

Even without CUDA-level throughput, an M4 Max with 48 GB unified memory and a 40-core GPU should be capable of much better iteration speed. Other diffusion pipelines (FLUX GGUF, SD3.5) run at reasonable speeds on this hardware.

## Possible Investigation Areas

- **Transformer forward pass**: The Qwen-Image transformer may have operations that don't map well to Metal (e.g., specific attention patterns, 3D RoPE, or dtype mismatches causing CPU fallback)
- **Memory pressure / thermal throttling**: If the text encoder isn't fully freed before denoising, unified memory contention could cause GPU throttle
- **Dtype overhead**: Check whether BF16 vs F32 matmuls are hitting fast Metal paths or falling back to emulation
- **Block-level offloading**: Verify offloading isn't being triggered unexpectedly on a 48 GB machine
- **Comparison needed**: Profile against CUDA on equivalent-class hardware to understand the Metal-specific gap

## Environment

- Branch: `fix/qwen-image-mps-regression` (includes quantized Qwen2.5-VL text encoder and CPU staging work)
- Model: qwen-image (GGUF quantized transformer)

jamesbrink:
Fixes #202.

## Summary
- stop dequantizing large quantized Qwen-Image transformer weights on every forward pass
- add device-local RoPE caching in the quantized transformer hot path
- route the quantized Metal attention path through SDPA with Metal-safe masking
- update the changelog

## Context
This is the already-pushed branch state only. Local follow-up experiments in `mold` and `../candle` were stashed and are not included in this PR.

### DIFF
--- a/.claude/skills/mold/SKILL.md
+++ b/.claude/skills/mold/SKILL.md
@@ -148,7 +148,7 @@ mold run ltx-video-0.9.6-distilled:bf16 "a waterfall" --frames 9 --format webp -
 
 **Constraints:** Frame count must be 8n+1 (9, 17, 25, 33, 49, ...). Dimensions must be multiples of 32. Current LTX defaults are 1216x704, 25 frames, 30 fps. Distilled models use fewer steps.
 
-**Current status:** `ltx-video-0.9.6-distilled:bf16` is the safest default. The `0.9.8` models require a spatial upscaler asset, which mold now pulls and resolves explicitly. mold currently runs the `0.9.8` first pass, but not the second multiscale refinement pass, so `0.9.8` is not yet at full upstream quality parity.
+**Current status:** `ltx-video-0.9.6-distilled:bf16` is still the safest default, but the `0.9.8` models now run the full multiscale refinement path. mold pulls the required spatial upscaler asset explicitly, keeps the shared T5 assets under `shared/flux/...`, and intentionally continues using the compatible `LTX-Video-0.9.5` VAE source until the newer VAE layout is ported.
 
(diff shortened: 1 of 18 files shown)


jamesbrink:
Addressed the Greptile P2s on the refreshed branch head `6bf6b2c`:
- made the Metal SDPA mask contiguous before calling `sdpa`
- documented the bounded-in-practice RoPE cache assumption
- removed the unused `QuantizedQwenImageTransformer2DModel::new(..., _device)` parameter and updated the call site

Focused validation still passes: `cargo test -p mold-ai-inference qwen_image_ -- --nocapture`

PR #247, opened 2026-04-19
feat(server,tui): serve cached video preview GIFs over the API (#242)

### LINKED ISSUE #242: feat(server,tui): expose cached video preview GIFs over the API for remote gallery animation
## Summary

Follow-up to #179 / #238. The new animated preview playback works for local TUI gallery entries (when the file itself is GIF/APNG/WebP, or when a sibling `.preview.gif` exists in `~/.mold/cache/previews/`), but **server-backed entries showing MP4 outputs stay on `Loading…` forever**.

## Repro

1. `mold serve` on a remote GPU host (e.g. `hal9000`).
2. Generate an LTX-2 / LTX-Video MP4 there.
3. From a client, run `mold tui` against `MOLD_HOST=http://hal9000:7680`, open Gallery → select the MP4.

Expected: animated GIF preview plays (or at least a still frame).
Actual: `Preview` pane reads `Loading…` indefinitely.

## Root cause

`crates/mold-tui/src/app.rs::load_gallery_preview` does this for server-backed entries:

```rust
let cache_path = crate::gallery_scan::image_cache_dir().join(&filename);
// fetch_and_cache_image hits GET /api/gallery/image/<filename>
// → for an MP4 this caches raw MP4 bytes
if let Ok(img) = image::open(&cache_path) {  // fails on MP4
    …
}
```

The cached `.preview.gif` lives on the server in `~/.mold/cache/previews/<name>.preview.gif` (written by the TUI when it generates a video, and by the server's video pipeline for its own outputs), but there is no HTTP endpoint to fetch it from. So the client falls through to decoding the MP4 as an image, which fails silently.

### Server (`mold-server`)

Add `GET /api/gallery/preview/:filename` that:
- Resolves `~/.mold/cache/previews/<filename>.preview.gif` (or whatever the server's `preview_dir()` resolves to).
- 200 with `Content-Type: image/gif` + the bytes when it exists.
- 404 when the source file isn't a video, or no preview was generated.
- Reuse the streaming + ETag helpers already in `routes.rs` for `/api/gallery/image/:filename`.

Optionally extend `GET /api/capabilities` to return `{ gallery: { has_preview_gifs: bool } }` so the SPA / TUI can probe support before requesting.

### Client (`mold-tui`)

In `load_gallery_preview` for server-backed entries, when the filename's extension is one of `mp4` / `webm` / etc., fetch `/api/gallery/preview/<filename>` first. On 200 → run the bytes through the existing `crate::animation::decode_animation_bytes` + `AnimationState` install path (same as the `BackgroundEvent::GalleryPreviewReady` branch). On 404 → fall back to fetching the full file (preserves today's behavior for raster outputs).

Cache the GIF locally under the existing image cache dir (or a sibling `previews/` subdir) so we don't re-fetch on every selection.

### Web SPA

Out of scope for this issue — the web gallery already plays MP4s directly via `<video>` and doesn't need the preview GIF. The endpoint should just be safe to call from the SPA if it ever wants a lightweight loop preview.

## Why not decode MP4 on the client?

`mold_inference::ltx2::media::extract_thumbnail` can pull the first frame, but it'd pull the whole MP4 over the wire for every preview selection, and only yields a still — no animation. Serving the already-generated GIF is faster and matches what the local path does.

## Acceptance

- [ ] `GET /api/gallery/preview/<filename>` returns the cached GIF for MP4 outputs and 404 for missing/non-video items.
- [ ] TUI gallery detail view animates remote video entries the same way it animates local ones.
- [ ] Regression test: server route returns 404 when no preview GIF exists; returns the bytes when it does.
- [ ] Regression test: TUI animation install path runs for server-backed video entries.

## Context

- PR that added local-side animation: #238
- Original issue: #179
- File pointers:
  - `crates/mold-tui/src/app.rs` (`load_gallery_preview`, ~line 2640)
  - `crates/mold-tui/src/animation.rs` (decoder + `AnimationState`)
  - `crates/mold-tui/src/thumbnails.rs` (`preview_gif_path`, `preview_dir`)
  - `crates/mold-server/src/routes.rs` (existing gallery handlers to mirror)

jamesbrink:
Closes #242.

## Summary

- `GET /api/gallery/preview/:filename` in `mold-server` streams `$MOLD_HOME/cache/previews/<filename>.preview.gif` as `image/gif` when present, returns 404 otherwise. Filename is sanitized (basename only) and honors `is_output_disabled()` like the sibling gallery routes.
- TUI's server-backed `load_gallery_preview` now fetches the preview endpoint first for video entries (`mp4`/`webm`/`mov`/`mkv`), caches the bytes under `image_cache_dir()`, and installs them through the existing `BackgroundEvent::GalleryPreviewReady` animation pipeline. On 404 it falls back to the raw-file fetch so raster outputs and preview-less videos still render.
- `MoldClient::get_gallery_preview` returns `Option<Vec<u8>>` so the 404 case is a normal signal rather than an error. New TUI helpers `preview_cache_path`, `fetch_and_cache_preview`, `is_video_filename` live in `gallery_scan`.

Before this change remote TUI gallery entries for MP4 outputs sat on `Loading…` forever because the client downloaded the raw MP4 via `/api/gallery/image/:filename` and `image::open` couldn't decode it.

## Test plan

- [x] `cargo fmt`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test -p mold-ai-server gallery_preview` — new route test (200 body + `image/gif` when preview present, 404 when missing) passes.
- [x] `cargo test -p mold-ai-tui gallery_scan` — new helper tests pass.
- [x] `cargo test --workspace` — only pre-existing flaky `device::tests::free_vram_returns_available_not_just_free_on_macos` regresses; passes on rerun in isolation (TOCTOU between two `macos_vm_stats()` syscalls, unrelated to this PR).
- [ ] Manual: run `mold serve` on a GPU host with an MP4 in the gallery + a matching `.preview.gif` under `$MOLD_HOME/cache/previews/`, then open `mold tui` against that host and confirm the detail pane animates the video entry.


jamesbrink:
Addressed both Codex findings in 55eb0ab:

**P1 — server never populated the preview cache.** `save_video_to_dir` now calls a new `save_video_preview_gif` helper that writes `video.gif_preview` to `$MOLD_HOME/cache/previews/<filename>.preview.gif` alongside the primary MP4, so the new endpoint returns real bytes for freshly generated videos instead of always 404'ing. Regression test in `crates/mold-server/src/queue.rs` pins the path layout.

**P2 — 404 fallback handed MP4 bytes to the image pipeline.** The TUI fallback now fetches the PNG thumbnail (which `/api/gallery/thumbnail/:filename` already generates via openh264 first-frame extraction for MP4s) instead of the raw video. Preview-less / older-server entries now show a decoded still instead of a blank pane.

jamesbrink:
Addressed both Codex findings in b0df8d6:

**P1 — multi-GPU path bypassed the sidecar save.** `save_video_preview_gif` is now `pub(crate)` and `gpu_worker::process_job` mirrors the single-GPU write, so `/api/gallery/preview/:filename` actually returns bytes on GPU-pool deployments (the primary target of this feature).

**P2 — orphaned sidecar leaked deleted content.** `get_gallery_preview` now requires the source MP4 under `output_dir` to still exist before serving the sidecar (404s otherwise), and `delete_gallery_image` removes the `.preview.gif` in lock-step with the thumbnail + metadata row. Regression test extended with an orphaned-sidecar fixture asserting the 404.

PR #270, opened 2026-04-24
fix(web): always-on delete, script-mode composer parity, tray-aware action bar

jdilley:
## Summary

Three UI/UX fixes to the web gallery:

- **Composer parity across modes.** The 🖼️ source-image picker and ⚙ settings icon buttons now render in a persistent toolbar above the Single/Script toggle, so they're visible in both modes. In Script mode, 🖼️ opens the picker targeted at stage 0 via a new `openStagePicker(i)` method exposed from `ScriptComposer`. The ✨ prompt enhancer remains per-stage inside each \`StageCard\`.
- **Delete is always on.** Dropped the \`MOLD_GALLERY_ALLOW_DELETE\` env gate: server always permits \`DELETE /api/gallery/image/:filename\`, \`/api/capabilities\` always returns \`{ gallery: { can_delete: true } }\`, and the SPA drops every \`canDelete\` prop/gate so the select + delete UI is always visible. TUI loopback \`mold serve\` spawn no longer sets the now-dead env var; the obsolete \`gallery_delete_toggle_reads_env_var\` and \`background_server_command_enables_gallery_delete\` tests are retired.
- **Bulk-select bar sits above the GPU/CPU tray.** \`ResourceTray\` publishes its live rendered height to \`:root\` as a \`--mold-tray-height\` CSS custom property via \`ResizeObserver\`. The selection action bar and back-to-top FAB in \`GalleryPage\` anchor their \`bottom\` to \`calc(var(--mold-tray-height, 0px) + max(0.75rem, env(safe-area-inset-bottom)))\`, so both move up when the tray expands and never get covered.

Pair \`MOLD_API_KEY\` when the server is exposed beyond localhost — that's the remaining guard on destructive routes.

## Test plan

- [x] \`cargo fmt --all -- --check\`
- [x] \`cargo clippy --workspace --all-targets -- -D warnings\`
- [x] \`cargo check -p mold-ai --features preview,discord,expand,tui,webp,mp4\`
- [x] \`cargo test --workspace\` (2,487 passed)
- [x] \`bun run fmt:check && bun run test && bun run build\` in \`web/\` (110 tests)
- [x] \`bun run verify && bun run build\` in \`website/\`
- [ ] Smoke test on <gpu-host>: pull + rebuild + restart \`mold-server\`, open gallery, confirm: image + settings buttons visible in both Single and Script modes; bulk select + delete work without env var; action bar floats above the tray both collapsed and expanded.


jdilley:
Cherry-picked into #271 as commit `1c7a6c7` (the catalog-expansion umbrella PR), since #271 also touches the same web/server surfaces and we want one open PR. Closing here. Test plan items from this PR are folded into #271's test plan.

Issue #286, opened 2026-05-17
mold-web bun install hangs on aarch64-darwin: bun2nix hook bypassed, makes live npm fetches


jamesbrink:
## Summary

On `aarch64-darwin`, `nix build .#mold-web` hangs indefinitely (observed: 18h+ before manual kill) during `bunNodeModulesInstallPhase`. The hook does not appear to be preventing bun from contacting the npm registry — `lsof` shows 50 active HTTPS connections to Cloudflare/npm during what should be an offline install. After the network phase, the build hits `EACCES: Permission denied (rename())` while bun tries to move packages into its global cache.

## Environment

- Host: macOS 15.4 (aarch64-darwin), Apple Silicon
- Nix: Determinate Nix
- mold ref: `bc1ca37` (current `main` HEAD as a flake input from `nixos` repo)
- Consuming flake: `inputs.mold.packages.\${system}.default` added to `environment.systemPackages` via a shared module
- Build path: `/private/tmp/nix-build-mold-web-0.9.0.drv-0` (Darwin builds aren't sandboxed by default)

### 1. Network access during bunNodeModulesInstallPhase

```text
$ sudo lsof -p <bun-pid> | grep -cE 'TCP|UDP|IPv'
50
$ sudo lsof -p <bun-pid> | grep TCP | head -3
bun ... IPv6 ... TCP [...:60995]->[2606:4700::6810:422]:https (ESTABLISHED)
bun ... IPv6 ... TCP [...:60911]->[2606:4700::6810:422]:https (ESTABLISHED)
bun ... IPv6 ... TCP [...:60893]->[2606:4700::6810:122]:https (ESTABLISHED)
```

`2606:4700::*` is Cloudflare; matches what `registry.npmjs.org` resolves to.

### 2. Indefinite hang

`bun install --linker=isolated --ignore-scripts` from one run kept running for **18h 41min** without any progress. No CPU activity, no log output. Killing the bun process unblocked the nix-daemon.

### 3. EACCES on cache moves (post-kill, retry attempt)

```text
mold-web> error: moving \"@types/sortablejs\" to cache dir failed: EACCES: Permission denied (rename())
mold-web>   From: .7ab1f2fefdc5ff69-00000030.sortablejs
mold-web>     To: @types/sortablejs@1.15.9@@@1
mold-web> error: moving \"@vue/test-utils\" to cache dir failed
mold-web>   From: .fab47eff9bbf33be-0000001F.test-utils
[...]

PR #287, opened 2026-05-18
fix(flake): fix mold-web bun install hang on aarch64-darwin + pure flake

jamesbrink:
## Summary

- **Root cause**: `bunInstallFlags = [ "--linker=isolated" ]` was overriding bun2nix's hook defaults, silently dropping `--backend=symlink`. Without it, bun ignores `BUN_INSTALL_CACHE_DIR` on macOS and makes live npm network fetches — which the Nix sandbox blocks, hanging indefinitely (issue #286).
- **Fix**: Remove explicit `bunInstallFlags`; rely on bun2nix defaults (`--linker=isolated --backend=symlink`).
- **Lifecycle scripts**: Add `dontRunLifecycleScripts = true` to skip the second `bun install` (without `--ignore-scripts`). All native-binary packages (esbuild, rollup, `@tailwindcss/oxide`, `lightningcss`) ship platform tarballs in `bunDeps` already; their postinstall scripts would attempt network access the sandbox blocks.
- **bun2nix bump**: v2.0.8 → v2.1.0. Drops `import-tree` transitive dep; `flake-parts`'s `nixpkgs-lib` now follows `["bun2nix", "nixpkgs"]` instead of a separate pinned `nixpkgs.lib` node. Lock cleaned up accordingly.
- **Version sync**: Bump `commonArgs.version`, `mold-web.version`, and `web/package.json` from `0.9.0`/`0.8.1` → `0.10.0` to match workspace `Cargo.toml`.

## Test plan

- [ ] `nix build .#mold-web` completes without hanging on aarch64-darwin
- [ ] `nix flake check` passes
- [ ] `nix build` (full mold binary) succeeds on aarch64-darwin and x86_64-linux
- [ ] `nix fmt` reports no changes

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,7 +3,6 @@
     "bun2nix": {
       "inputs": {
         "flake-parts": "flake-parts",
-        "import-tree": "import-tree",
         "nixpkgs": [
           "nixpkgs"
         ],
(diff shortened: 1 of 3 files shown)


jamesbrink:
closes #286

PR #375, opened 2026-07-12
feat: server queue controls + desktop Jobs view, run History, dock badge fix, local saves

jamesbrink:
## Summary

Part 3 of 4 (stacked on #374).

### Server (mold-server + mold-core)
- `POST /api/queue/pause` / `POST /api/queue/resume` — gate both dispatch loops (single-GPU worker + multi-GPU dispatcher) via a `QueuePause` (AtomicBool + Notify with lost-wakeup protection); the running job always finishes; idempotent, `{"paused": bool}`.
- `DELETE /api/queue` — cancel-all-queued in one write-lock sweep, one `job_ended` per job, `{"cancelled": n}`; running jobs untouched.
- `/api/status` gains `queue_paused`; `/api/capabilities` gains `queue: {can_pause, can_cancel_all}`; `/api/events` emits `queue_paused`/`queue_resumed` on real transitions only.

### Desktop
- **Jobs view (⌘6)**: per-host sections of the full server queue (other clients' rows marked `OTHER CLIENT`), live thumbnails/step progress for our jobs, per-job cancel, Pause/Resume + two-step Cancel all (capability-gated — hidden on older servers), finished-this-session list with one-click reuse. Pause state stays live across clients via the new events.
- **History**: opens on a **Runs** tab — day-grouped log of every finished generation (thumbnail, model, WxH, seed, steps) with full-settings reuse (seed included); the prompt log remains as **Prompts**.
- **Dock badge**: now counts THIS app's active jobs, event-driven; previously mirrored the whole engine queue depth on a 10s poll (confusing on shared/remote engines).
- **Local saves of remote prints**: completed remote/RunPod generations are written into the local output dir automatically (Settings → App toggle, default on) — bytes come straight from the SSE complete event (works through RunPod proxies, embedded metadata intact) with an immediate local-gallery DB row via `mold_db::persist`. Gallery right-click gains **Save to this Mac**.

## Test plan
- 414 desktop vitest tests pass (new: jobs store, JobsView, HistoryView runs, dockBadge, auto-save routing incl. pref-off/local-job cases, queue-pause event sync)
- 495 mold-server tests + mold-core tests pass (new: QueuePause blocking/wakeup, cancel_all_queued invariants, route tests, serde wire shapes)
- 50 src-tauri tests pass (new: unique_output_path collisions, save_remote_outputs defaults)
- `cargo clippy --workspace --all-targets` clean (incl. TUI/discord ServerStatus fixture updates), `cargo fmt --check` clean, vue-tsc + prettier clean


jamesbrink:
Post-review notes (adversarial review workflow ran over the full stack; 21 findings confirmed, all fixed in follow-up commits except these deliberate deferrals):

1. **No Keychain→file migration** (flagged HIGH by review): intentional — the fresh-start behavior was an explicit product decision (re-enter keys once; zero Keychain prompts ever). Documented in the CHANGELOG entry.
2. **Cancel-all while paused leaves queue_depth inflated until resume**: cancelled jobs stay in the dispatch channel and are skipped (closed result channel) when the dispatcher drains on resume — identical semantics to the pre-existing per-job `DELETE /api/queue/:id`. No deadlock: resume always drains. A counter-decrement-at-cancel refactor would touch queue-counter ownership and is left for a dedicated change if the lag ever matters in practice.
3. **`host_id` slug collisions** (e.g. `http://` vs `https://` same host:port map to one slug): accepted — colliding hosts share one saved-host entry and API-key slot, which for the same physical server is arguably correct. Changing the slug format now would orphan keys persisted by this stack.

PR #390, opened 2026-07-13
feat(desktop): per-host models, model-aware routing, and a most-capable host option

jamesbrink:
## Summary

Multi-host generation gets model intelligence (user request: "use models on a host that might not be available locally… and an option to run on the most powerful/capable host"):

- **Per-host model availability**: new `hostModels` store fans `GET /api/models` out to every ready host (60 s staleness, per-host failure isolation). The Generate model picker now shows the **union** across hosts with availability tags (host label / "N hosts"), and a "not on <host> — will download there" hint when the sticky target lacks the pick. Selecting a model that only exists on an extra host works (defaults populate from the union entry).
- **Model-aware Auto routing**: `resolveRoute(selection, modelName)` restricts least-busy Auto to hosts that have the model (falls back to all ready hosts + the server's auto-pull when none do).
- **"Most capable" routing option** in the Host selector: backend class (CUDA > Metal > unknown) → total VRAM → queue depth → local on full ties; model-aware like Auto. Persisted through the existing `generateTargetHost` setting via a `"capable"` sentinel (degrades to Auto on older builds).
- **Additive wire change**: `gpu_info.backend` on `/api/status` (serde-default, omitted when unknown; older peers unaffected — back-compat tests). Client detection ladder: `gpu_info.backend` → GPU-name inference. Bonus: macOS servers now report a Metal `gpu_info` block (unified memory as VRAM) instead of `null`, so Metal hosts rank correctly.

(Replaces #389 — same content, rebased onto current main.)

## Test plan (failing-first)

- `inferBackendFromGpuName` / `backendRank` tables; `pickMostCapableHost` (cuda beats metal despite deeper queue, VRAM/queue tiebreaks, model filter + fallback, nothing-ready → null); `resolveRoute("capable")` + model-aware Auto; `hostModels` (fan-out to ready hosts only, dead-host keeps last entries, staleness/force, union dedup, `hostsFor`); HostSelector renders Most capable; Rust `GpuInfo.backend` serde back-compat.
- Gates: `cargo fmt --check` ✓ · `clippy --workspace -D warnings` ✓ · `cargo test -p mold-ai-core -p mold-ai-server` ✓ · desktop `fmt:check`/`test` (507)/`build` ✓ (all re-run after the rebase onto main past #386)


jamesbrink:
Independent peer review: **no blockers** — routing helpers, fallbacks, and the additive wire change all verified (fresh-boot capable-pick, pre-fetch Auto, explicit-pick semantics, all nine `GpuInfo` construction sites, Metal/CUDA unit consistency). Applied the one gating item in 8f2490a3: the ready-set watch is now `immediate` and picker-open force-refreshes, so model-aware routing engages from the FIRST Generate click and cross-client downloads on extra hosts surface when the user looks. Sentinel comment softened per review. Deferred (reviewer-acceptable): direct component tests for pickerFamilies/availabilityTag wiring, web-SPA GpuInfo type parity.

PR #481, opened 2026-07-21
feat(core): queue listing and cancel client methods

jamesbrink:
## Summary

Part of the G13 TUI-redesign PR series — prerequisite for the TUI Machines workspace, which needs typed Rust access to the server queue that so far only the web SPA consumed raw.

Wraps two existing server endpoints in `mold_core::MoldClient`:

- `list_queue()` → `GET /api/queue`
- `cancel_queue_job(id)` → `DELETE /api/queue/:id`

backed by new Deserialize-capable twins of mold-server's Serialize-only `job_registry` types:

- **`QueueJobEntryWire`** — `id`, `model`, `state`, `started_at_unix_ms`, `position`, plus additive `gpu`, `target_gpu`, `seed_pinned`, `metadata` (`Option<Box<OutputMetadata>>`).
- **`QueueListingWire`** — the `{ "entries": [...] }` wrapper the server actually returns (`QueueListing` wraps the rows in an object precisely so the response can grow fields; this is not a bare array).

## Forward-compat decisions

- `state` is a plain `String` (`"queued"` / `"running"` today) — a future server lifecycle state deserializes cleanly instead of breaking old clients.
- All additive fields are `#[serde(default)]` so older servers (pre `seed_pinned`/`metadata`) parse, and `skip_serializing_if = "Option::is_none"` so re-serialization matches the server's contract exactly (queued rows carry no `gpu` key at all — never `"gpu": null`).
- Unknown extra fields on both the wrapper and the rows are ignored (serde default), matching the server's stated intent to grow the wrapper.
- `cancel_queue_job` routes non-2xx through the existing `error_for_status_with_body` helper so the 409 already-running reason (and 404 body) reach the caller verbatim; the id is percent-encoded via the existing `encode_path_segment`. `list_queue` follows the `server_status()` request/error pattern; both inherit the client's `x-api-key` default-header mechanism.

Purely additive — no behavior change for existing consumers. No other crates touched.

## Test plan

TDD — tests written first and watched fail, then implemented:

- Serde fixture deserializing a realistic `GET /api/queue` body mirroring `job_registry.rs`'s own serialization tests (running row with `gpu`, queued row with `target_gpu`/`seed_pinned`/`metadata`, unknown future fields on wrapper + row).
- Unknown-state round-trip (`"state": "some-future-state"`).
- Missing-optional-fields fixture (older server shipping only the five original fields).
- Absent options stay off the wire on re-serialization; empty `{}` wrapper defaults to no entries.
- wiremock (existing client.rs pattern): wrapped-listing parse, `x-api-key` header sent, 204 cancel success, 409 error message contains status + body text.

`cargo test -p mold-ai-core` green (847 lib tests), `cargo clippy -p mold-ai-core --all-targets -- -D warnings` clean, `cargo fmt --all -- --check` clean. CHANGELOG `[Unreleased]` → `### Added` updated.


jamesbrink:
Superseded: #488 (Machines workspace) had merged this branch as its prerequisite, and its squash carried the entire queue-client changeset (QueueListingWire/QueueJobEntryWire wire types, MoldClient::list_queue/cancel_queue_job, all tests) and the CHANGELOG bullet to main verbatim — both verified present on main. Closing as redundant; no content was lost.

PR #527, opened 2026-07-24
feat(models): kind + sort filters on desktop and iPhone Discover

jamesbrink:
## Summary

Brings desktop and iPhone Discover to parity with the web catalog's filter set:

- **Kind chips** (Models / LoRAs / CLIP / Text encoders / VAEs / Tokenizers / ControlNet) on desktop `CatalogTab` and the iPhone catalog view — option lists shared via new `desktop/src/lib/catalogFilters.ts`, labels verbatim from the web topbar.
- **Sort selector** (Downloads / Rating / Recent) on both surfaces; `sort` param added to the shared `CatalogSearchParams` request layer. Defaults (`downloads`, All kinds) are omitted from the wire, mirroring the `source: "all"` pattern.
- Filter changes reset pagination to page 1 via the existing search watchers; manifest/installed rows are hidden client-side under a non-checkpoint kind; iPhone kind chips are a horizontally scrollable 44pt row, hidden in the Installed segment (kind clears entering Installed, like family).
- No persistence — matches the existing ephemeral-filter pattern.

## Test plan

- TDD: 7 new tests written red-first — `catalogFilters.test.ts` pins option values inside the server wire vocabulary (guards the 422 on unknown sort), URL-assertion tests in `api/catalog.test.ts`, component tests for page-1 reset + row hiding in `CatalogTab.test.ts` / `MobileCatalogView.test.ts`.
- `bun run check:frontend` green (studio 23, web 784, desktop 1967 tests + both builds); `bun run build:mobile` green; website docs gate green.

Docs: CHANGELOG, website desktop/iphone guides, apps/mobile/README, desktop/docs/feature-parity.md, mold SKILL.md.

https://claude.ai/code/session_01VQkSfCvSD58mgWE1Gt1E1J


jamesbrink:
Codex peer review flagged one P2: when browsing a remote host old enough to predate catalog sorting, a non-default sort is sent but silently ignored (download order returned). Decision: accepted as a known limitation rather than adding capability negotiation — the web catalog UI has always sent sort unconditionally with the same behavior against old servers, and the server-side sort vocabulary predates every host version currently supported by the desktop app's wire-compat floor. Revisit if /api/capabilities grows a catalog block.

PR #545, opened 2026-07-26
fix(web): allow catalog credential overrides

jamesbrink:
## Summary

- make web-saved Hugging Face and Civitai tokens override server environment defaults
- let clearing a saved override immediately reveal and reuse the environment fallback
- keep credential status masked while labeling Environment versus Saved override in web Settings
- add a symmetric `services.mold.civitaiTokenFile` NixOS option and document both token-file values as overridable defaults
- synchronize the changelog, README, canonical project guidance, API docs, catalog docs, remote workflow guidance, and NixOS deployment docs

## Root cause

The server resolver selected `HF_TOKEN` / `CIVITAI_TOKEN` before the owner-only saved credential and rejected PUT/DELETE requests whenever an environment token existed. That made environment-provided values immutable in the web interface instead of usable as defaults.

## Validation

- `nix develop -c cargo test -p mold-ai-server catalog_credentials::tests:: --lib` (7 passed)
- `bun run --cwd web test` (859 passed)
- `bun run --cwd web build`
- `bun run --cwd web fmt:check`
- `bun run --cwd website fmt:check`
- `bun run --cwd website verify`
- `bun run --cwd website build`
- `nix fmt nix/module.nix`
- evaluated `services.mold.civitaiTokenFile` through `nixpkgs.lib.nixosSystem`


jamesbrink:
Addressed the low-confidence Copilot summary note in 1038110b: credential status now trims and filters saved/environment values exactly like runtime resolution, with a regression test proving a whitespace-only saved value reports and uses the environment fallback.

Issue #600, opened 2026-07-29
tui: LTX-2 file inputs (audio/source video/keyframes/retake), camera preset, and chain-jobs view


jdilley:
The TUI hardcodes every LTX-2 advanced field to `None` when building requests (`crates/mold-tui/src/backend.rs` request construction, as of 1503696d): `enable_audio`, `audio_file`, `source_video`, `keyframes`, `pipeline`, `loras`, `retake_range`, `spatial_upscale`, `temporal_upscale`. The param panel exposes only the image-era fields plus Frames/Fps. ~~The TUI script composer covers chain prompts/frames but there is no durable chain-job management view.~~ The TUI script composer was deleted by #1619 — sequence authoring is CLI/API-only now — so that half is out of scope by design; a read-only chain-jobs *view* mirroring `mold jobs` is still wanted.

**Scope**
- Add param fields for: audio on/off, pipeline mode, spatial/temporal upscale, camera preset; file-picker-style fields for audio file / source video / keyframes where feasible in a TUI.
- A jobs view for `/api/chain-jobs` (list/show/resume/retake/cancel) mirroring `mold jobs`.
- Keep the layout-constant regression tests in step (inner-area-fits-rows invariants).

_Audit 2026-09-10 against `main` @ `fdb3225c`:_ roughly half shipped. Landed: audio tri-state, pipeline mode, spatial/temporal upscale, guidance overrides. Still open: `crates/mold-tui/src/backend.rs:1706-1717` hardcodes `audio_file`, `audio_file_path`, `source_video`, `source_video_path`, `keyframes`, `ic_lora_control`, `loras`, `retake_range` to `None`; zero hits for `camera_preset` in `crates/mold-tui`; no chain-jobs view. Demilestoned from Surface Parity (closed 2026-09-10) as a long-tail TUI item.

jamesbrink:
Status audit against current main (`3bdbf4b2`) — this issue is about half delivered, so narrowing the live scope:

**Shipped** (TUI Create form's Advanced accordion, `crates/mold-tui/src/ui/create_form.rs` + `backend.rs`):
- `enable_audio` tri-state (default/on/off, MP4 auto-select, hidden when `supports_audio:false`)
- `pipeline` mode selection (preserves server default until touched)
- `spatial_upscale` (native/1.5x/2x) and `temporal_upscale` (native/2x)
- `guidance_overrides` (STG scale/blocks, CFG rescale, modality scale, guidance skip), all absent-until-touched with regression tests

[...]

PR #643, opened 2026-07-31
feat(studio): add native LTX-2 guidance controls

jamesbrink:
## Discovered parity gap

PR #642 added the additive LTX-2 `guidance_overrides` contract plus CLI flags and web controls, but the desktop and iPhone Create surfaces could neither author nor restore those settings. A saved print or request could therefore retain the server metadata while the native forms silently dropped it on reuse.

## Before / after parity matrix

| Surface | Before | After |
| --- | --- | --- |
| CLI | Supports all five overrides | Unchanged |
| Web | Supports all five overrides | Unchanged |
| Desktop | Missing controls and native form persistence | LTX-2 Advanced controls, validation, counts/reset, request + metadata/template restoration |
| iPhone | Missing controls and native form persistence | Same contract in the Advanced sheet with mobile layout, validation, counts/reset, and restoration |
| TUI | Uses pipeline defaults | Intentionally unchanged; tracked as part of the broader advanced-video parity gap in #600 |

## Acceptance criteria

- Untouched native controls omit `guidance_overrides`, preserving checkpoint pipeline constants.
- Desktop and iPhone expose STG scale/blocks, CFG rescale, modality scale, and guidance skip stride only for LTX-2.
- Invalid values are named before submission and block the request; valid values serialize through the shared Studio contract.
- Advanced badges count the guidance group once, group/full resets clear it, and saved metadata/templates restore it.
- Non-LTX-2 requests are pruned, and automatic long-video chaining refuses to silently discard overrides.

## Implementation

- Extended the shared native request/metadata form types and reused `studio/lib/guidanceOverrides.ts` for parsing, serialization, bounds, and restoration.
- Added desktop and iPhone controls with responsive native layouts, accessible labels/error alerts, active counts, and reset behavior.
- Added legacy-template hydration, request/Library metadata restoration, family pruning, submit validation, and chain-loss detection.
- Updated the feature matrix, desktop/iPhone guides, maintainer instructions, README, and changelog.

## Intentional non-parity

- TUI remains on pipeline defaults because its advanced-video controls are a separate, already tracked capability gap (#600); this PR does not introduce shell-local contract logic.
- Durable sequences/chained long video do not yet carry guidance overrides. Native Create now blocks that lossy route and tells the user to remove the overrides or reduce the frame count instead of pretending the settings survived.

## Regression coverage

- Form defaults, cloning, family reconciliation, serialization, request/metadata reuse, and legacy templates.
- Desktop and iPhone rendering, editing, validation, active counts, and resets.
- Capability pruning and automatic-chain incompatibility detection.

## Validation

- `bun run --cwd desktop test` — 261 files / 2,725 tests passed
- `bun run build:desktop`
- `bun run build:mobile`
- `bun run check:architecture`
- `bun run check:dead-code`
- `bun run --cwd desktop fmt:check`
- `bun run --cwd website fmt:check`
- `bun run --cwd website verify`
- `bun run --cwd website build`
- `bash scripts/tests/ios-release-assets.sh`
- `nix develop -c desktop-check`
- `nix develop -c ios-check`
- `nix develop -c ./scripts/ios.sh simulator` — signed simulator bundle verified
- `(cd desktop && nix develop -c cargo tauri build --debug --bundles app --features metal)`
- `git diff --check`

### Desktop

Ran the packaged debug `Mold - dev` app against a temporary reachable LTX-2 fixture host. Selected the LTX-2 checkpoint and observed all five Guidance overrides controls. Entering STG scale `1.5` and invalid blocks `28, nope` produced the field-named inline error, a two-field group count / one-group Advanced badge, and disabled Generate. Correcting the list to `28, 29` removed the error and enabled Generate. Guidance Reset cleared the fields and counts. The fixture was then forgotten and This Mac restored as the generation target.

### iPhone

Built, signed, installed, and launched Mold on the booted iPhone 17 / iOS 26.0 simulator against the reachable `plato` host. Selected `ltx-2-19b-distilled:fp8`, opened the full-screen Advanced sheet, and observed the same five controls with 44pt mobile fields. Invalid STG blocks showed inline feedback and the group/Advanced counts; correcting to `28, 29` cleared the error, and Reset advanced restored every override placeholder and removed the badges. The prior host and Sequence output mode were restored, then the app was stopped.

CLI and web behavior was unchanged and covered by the existing contract/tests; TUI was intentionally not modified.

## Remaining risk

The only known parity gap in this bounded area is TUI advanced-video authoring, tracked separately in #600. Chain schemas still lack these overrides, but this PR fails closed before queueing rather than losing user intent.


desktop/src/mobile/mobile.css
```diff
@@ -1625,6 +1625,26 @@ button.mobile-generate-disclosure {
...
+  display: inline-flex;
+  min-width: 22px;
+  min-height: 22px;
+  align-items: center;
+  justify-content: center;
+  margin-left: 6px;
+  padding: 0 6px;
+  border: 1px solid var(--control-edge);
+  border-radius: var(--radius-pill);
+  font-family: var(--font-mono);
```
jdilley: `--font-mono` is not defined in the mobile bundle — the iPhone entry (`desktop/src/mobile/main.ts` → `../styles/base.css` → `desktop/src/styles/tokens.css` + `ui/`) never loads `web/src/style.css`, which is the only place that alias exists. The badge silently falls back to the inherited body font while the desktop twin in `AdvancedSettings.vue` correctly uses `var(--f-mono)`. Every sibling rule in this file uses `var(--font-utility)` (73 occurrences), which resolves to the same Martian Mono stack.

jdilley:
### Code review

Found 1 issue:

1. The new iPhone guidance count badge uses `var(--font-mono)`, which is undefined in the mobile bundle — that alias only exists in `web/src/style.css`, which the mobile entry never loads, so the badge silently falls back to the inherited body font. Every sibling rule in `mobile.css` uses `var(--font-utility)`, and the desktop twin uses `var(--f-mono)`. (Also flagged inline.)

https://github.com/utensils/mold/blob/70e5b9491bf3f12178232f8838145f3ca0bbc643/desktop/src/mobile/mobile.css#L1643-L1645

Also verified and found sound: wire field names round-trip between `guidanceOverridesToWire`/`FromWire` and the request/metadata types; bounds mirror `crates/mold-core/src/validation.rs`; `pruneRequestForFamily`, `reconcileModelCapabilities`, and both reset paths clear the state; legacy-template hydration covers both storage keys; both submit gates route `guidanceOverridesError`; sequence mode is correctly unreachable on both native surfaces; and the desktop/mobile auto-chain block is warranted (the chain wire hard-codes `guidance_overrides: None` per stage).

Lower-confidence observations, take or leave: `advancedActiveCount` in `MobileApp.vue` counts guidance overrides without the `supportsAdvancedVideo` gate that `advancedCount.ts` applies, so a Library reuse onto a substituted non-video model can leave a phantom badge count (`reuse.ts` clears the other advanced fields but not this one); the CHANGELOG line "CLI and Studio refuse to pretend an override survived a chain route" overstates web, which has no chain-loss guard for this field; `desktop/docs/feature-parity.md` lines 65/178 no longer match the desktop Video section and Reuse-settings behavior; and the `capabilities.test.ts` prune test wasn't extended for the new field.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

desktop/src/lib/chainRouting.ts
```diff
@@ -76,6 +77,9 @@ export function unsupportedAutoChainFields(req: GenerateRequest): AutoChainUnsup
   if (req.retake_range) unsupported.push("retake_range");
   if (req.spatial_upscale) unsupported.push("spatial_upscale");
   if (req.temporal_upscale) unsupported.push("temporal_upscale");
+  if (req.guidance_overrides && Object.keys(req.guidance_overrides).length > 0) {
+    unsupported.push("guidance_overrides");
+  }
```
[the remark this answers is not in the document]
jamesbrink: Fixed in d4ea13f5. `unsupportedAutoChainFields` now reports guidance loss only when at least one scalar is non-null or the block list is non-empty; an explicit zero remains meaningful. Added regression coverage for a null-only/empty-list object and for zero.

desktop/src/mobile/mobile.css
jamesbrink: Fixed in d4ea13f5. The mobile guidance badge now uses the defined `--font-utility` token, with a static CSS regression test matching the rest of the mobile shell.

jamesbrink:
Addressed the remaining review observations in d8a5a00b:

- capability-gated the iPhone Advanced badge so stale LTX-2 guidance cannot appear on a fallback family;
- clear guidance overrides when Library reuse substitutes a model whose family does not support advanced video, while preserving them for compatible LTX-2 replacements;
- extended `pruneRequestForFamily` coverage to prove non-LTX-2 stripping and LTX-2 preservation;
- corrected the changelog's chain behavior wording and updated the native feature inventory for the request field, Video controls, and desktop/iPhone reuse.

Regression-first focused coverage is green (144 tests), followed by the full 2,727-test desktop suite, desktop/mobile builds, formatting, and website verify/build. Thanks for the careful pass.

Issue #679, opened 2026-08-03
test: the execution_plan artifact-identity tests are racy on fast filesystems


jamesbrink:
## Summary

A cluster of ~5 tests in `crates/mold-server/src/execution_plan.rs` fails non-deterministically on a fast local machine while passing in CI. The failing *subset* varies run to run — I have observed 2, 3, 4, and 5 failures from the same cluster within a single session, with no source change between runs:

- `cache_miss_identity_changes_when_artifact_metadata_changes`
- `same_size_in_place_overwrite_changes_equivalence_content_identity`
- `same_size_in_place_overwrite_with_restored_mtime_changes_content_identity`
- `frozen_chain_model_uses_canonical_companions_and_ignores_changed_config`

## Root cause

`artifact_metadata_identity` derives identity from path + size + mtime. The fixtures overwrite a file **with different content of identical size**:

```rust
std::fs::write(&path, b"aaaa").unwrap();
let first_metadata = artifact_metadata_identity(&path, &path.metadata().unwrap());
// ...
std::fs::write(&path, b"bbbb").unwrap();
let second_metadata = artifact_metadata_identity(&path, &path.metadata().unwrap());
assert_ne!(
    first_metadata, second_metadata,
    "fixture must observe a new metadata identity"
);
```

Size is unchanged by construction, so the assertion rests entirely on the two writes landing in **different filesystem mtime ticks**. When they land in the same tick the identities are equal and the fixture's own precondition fails — before the test reaches the behaviour it means to check.

This is why it fails *more* on a fast machine: the two writes complete closer together. CI's slower runners and filesystem separate them, which is why `main` is green.

## Why it matters beyond flakiness

The failing set drifts with machine load, so a local `cargo test --workspace` cannot distinguish "my change broke something" from "the box was busy". Every contributor running the documented CI-equivalent gate has to hold a mental list of which failures to ignore, and that list is not even stable. During recent LTX-2 milestone work this cost real time on several PRs, each time re-deriving that the failures were unrelated.

## Suggested fix

Make the fixture control the clock instead of racing it — force a distinct mtime rather than hoping for one:

```rust
filetime::set_file_mtime(&path, FileTime::from_unix_time(t + 1, 0))?;
```

`filetime` is already in the dependency graph. Alternatively, assert on the content identity these tests actually care about and drop the metadata-identity precondition, which is an implementation detail of the cache rather than the property under test.

Either way the tests then verify the intended invariant deterministically, on any filesystem.

## Reproduction

[...]

PR #681, opened 2026-08-03
feat(ltx2): HDR IC-LoRA output with OpenEXR export

### LINKED ISSUE #594: feat(ltx2): HDR IC-LoRA pipeline with EXR export
Upstream's `HDRICLoraPipeline` produces linear HDR float output for EXR export (VFX/compositing workflows). mold has no HDR path: all decode targets 8-bit SDR formats.

**Scope**
- HDR IC-LoRA adapter support on the existing `IcLora` mode.
- Linear float output path through the VAE decode + an EXR (or at minimum 16-bit PNG) writer; likely `--spatial-tile`-style chunking for VRAM.
- CLI/web flag + docs. Gallery can store a tonemapped SDR preview alongside the EXR.

Lower priority than control LoRAs / long-form, but it is a capability upstream advertises that mold cannot express at all today.

jamesbrink:
Closes #594.

## What this does

The LTX-2.3 HDR IC-LoRA emits an ARRI **LogC3**-encoded signal. Mold collapsed that to 8-bit like any other render, throwing away exactly the range the adapter exists to produce — the adapter worked, and its output was unusable for the grading it was built for.

`mold run --ic-lora-control hdr --hdr-exr-dir <dir>` now also writes one EXR per frame in scene-referred linear light, alongside the ordinary video. `--hdr-exr-full-float` selects 32-bit samples over the 16-bit default.

## Decisions worth reviewing

**The decode forks at the tensor, not after the frames.** The float→8-bit conversion is lossy and HDR cannot be recovered from the result, so `decoded_video_to_hdr_frames` is a sibling of `decoded_video_to_frames` rather than a post-process. Everything downstream of the SDR path is untouched.

**LogC3's toe is preserved unclamped.** Pure black decodes to `-0.0173`, not `0`, because upstream's `decompress` does not clamp its output (`ltx-core/hdr.py:43-49`). Clamping would silently alter every shadow in an export. This is documented in the model guide so a compositor seeing negative values knows they are intentional.

**EXR is rejected without `ic_lora_control=hdr`.** Applying the inverse transfer function to an ordinary SDR render produces a file that opens fine, has plausible values, and is wrongly graded — indistinguishable from intent. Refusing beats guessing. The rejection fires before the checkpoint loads.

**The sequence is a sidecar, not a gallery artifact.** A frame sequence is many files and gigabytes; `scan_output_dir` is depth-1 files-only and `generations` keys on one filename. The gallery artifact stays the ordinary tonemapped video, and `hdr_exr_dir` / frame count are recorded in `metadata_json`. This matches upstream and what the issue asked for.

**The adapter's shipped scene embeddings are now used instead of prompt encoding.** The HDR repository ships `ltx-2.3-22b-ic-lora-hdr-scene-emb.safetensors`, which mold downloaded and never read. Upstream takes it as a **required** argument and never encodes a prompt at all (`hdr_ic_lora.py:250-256`, `:383`) — the adapter was trained against that one fixed scene context, so encoding the user's prompt through Gemma was quietly running it out of distribution in a way the output would not reveal. Loading the companion also removes the Gemma encode from the critical path.

Consequence users are told plainly in the changelog and the model guide: **prompt text is ignored for this control.**

**Refusal over a type-satisfying lie.** Saved embeddings are one *conditional* context. Duplicating them into `NativePromptEncoding.unconditional` would compile and would cancel classifier-free guidance to nothing while still appearing to run, so `prepare` refuses any plan that would read that slot and names the reason. Upstream never reaches it: these embeddings ship for distilled checkpoints, which run at guidance 1.0.

**`exr` over the alternatives.** The pure-Rust `exr` crate is already vendored transitively via `image`, so this adds **zero** new dependencies and nothing to cross-compile across the sm86/89/100/120 matrix. `openexr` would pull a C++/cmake closure; `image::codecs::openexr` exposes no compression or attribute API.

## Verification

Real GPU renders on `ltx-2.3-22b-distilled:fp8` (RTX 4090, forced-local, seed 424242, 704x448, 25 frames, real weights):

| | Prompt-encoded | Scene embeddings |
|---|---|---|
| Wall time | 178.7 s | **120.0 s** |
| Path | `real path pipeline=IcLora` | same, plus `prepare.scene_embeddings 0.030s` |
| EXR min / max | −0.015839 / 37.593750 | −0.008560 / 26.593750 |
| Samples below 0 | 3,382 | 5,937 |
| Samples above 1.0 | 184,759 (19.5%) | 143,989 (15.2%) |

25 files, `frame_00000.exr`…`frame_00024.exr`, `gbrpf16le`, linear, sRGB/D65 chromaticities, ZIP16. `ffprobe` on the video matches the request exactly (704x448, 25 frames, 24 fps).

A caution for anyone re-checking these numbers: **`ffmpeg -pix_fmt gbrpf32le` clamps float output to `[0,1]` in swscale**, which reports a perfectly plausible `[0.000000, 1.000000]` and hides the entire HDR range. The values above were read with the `exr` crate the writer itself uses.

Guard rails confirmed end to end: `--hdr-exr-dir` without the HDR control errors before the checkpoint loads and creates no directory; `--hdr-exr-full-float` without a directory is rejected at argument-parse time.

Gates: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo check -p mold-ai --features preview,discord,expand,tui,webp,mp4` all clean. `cargo test --workspace` shows no new failures — the only failures are the pre-existing load-sensitive `execution_plan` cluster tracked in #679, which this PR does not touch.


jamesbrink:
Rebased onto `main` now that #672 has landed. The duplicated `feat(ltx2): register the gated lip-dub and HDR control adapters` commit is dropped — I diffed it against `main` first and confirmed nothing in it was missing before skipping, since `main` carries that registry plus the two review fixes from #672 on top. This branch is now only the seven HDR commits, and it builds on the merged `control_missing_files` shape rather than the pre-review one.

Also fixed the `docs` gate. Worth recording for anyone else touching `website/`: the repo root pins prettier **3.9.4** but the `website/` workspace resolves **3.8.1**, and CI runs 3.8.1. They disagree on emphasis markers (`*x*` vs `_x_`) and on table alignment for wide unicode glyphs, so `bun run fmt:check` from `website/` gives the wrong answer locally *in both directions* — on a clean `main` it reports `guide/tui.md` and `models/ltx2.md` as unformatted when CI considers them correct. The reliable local check is:

```bash
cd website && bunx prettier@3.8.1 --check .
```

Full gate re-run after the rebase: fmt and clippy clean, 1244 tests passing. The only failures are the pre-existing load-sensitive `execution_plan` cluster tracked in #679, which this PR does not touch — and whose failing subset visibly shifted between runs during this session, which is what that issue is about.

jamesbrink:
All three review findings addressed in `25f97284`. **Please do not merge yet** — the memory fix changed the write path materially, so the earlier GPU verification no longer covers it. Re-run is queued behind two agents on the shared 4090; I will post the numbers here before this goes in.

**1. Memory — the serious one.** You were right, and the arithmetic is worse than "very large":

| Shape | Buffered | Streamed |
| --- | --- | --- |
| 704x448 x 25 | 0.09 GB | 3.8 MB |
| 1920x1088 x 121 | **3.0 GB** | 25 MB |
| 1920x1088 x 481 (LTX-2's 20 s ceiling) | **12.1 GB** | 25 MB |

Frames are now converted and written one at a time inside the decode, so peak is one frame regardless of clip length. `DecodedVideo` and `NativeRenderedVideo` carry a count instead of the pixels, and `write_exr_sequence` is `#[cfg(test)]` — production streams via `write_exr_frame`.

Worth saying plainly: **my verification could not have caught this.** I chose 25 frames at 704x448 to keep the GPU-lock hold short, which is reasonable for testing correctness and useless for testing resource scaling. Correctness scales down; memory does not. The re-run will use a frame count that actually exercises it.

**2. Metadata — the PR body was wrong.** It claimed `hdr_exr_dir` was recorded and it was not. `OutputMetadata` now carries `hdr_exr_dir` and `hdr_exr_full_float`, populated in `from_generate_request`, with a round-trip test. Both are `skip_serializing_if`, and a test asserts an ordinary render's JSON contains no `hdr_exr` key at all, so existing rows and older readers see exactly what they saw before. Thank you for catching a documentation claim the code did not honour — that is the kind of gap that survives review indefinitely.

**3. Control-id normalization — fixed**, now routed through the shared `normalize_control_id` rather than a raw `trim()`.

One correction on that last point, since I tested your examples rather than assuming: **`hdr_` is not a valid alias.** The normalizer maps `_` to `-`, so `hdr_` resolves to `hdr-`, which is not a registered adapter id anywhere in the stack — it would fail identically before and after this change. I had copied that example straight into a test and it failed. The test now covers the real aliases (case and surrounding whitespace) with a comment explaining why a trailing underscore is not one, since the obvious reading is that it should be.

Gates after the change: fmt, clippy `-D warnings`, and the feature-combo check all clean; 1243 tests pass. The only failures are the #679 `execution_plan` cluster, which #682 fixes and this branch does not touch.

jamesbrink:
## A bug the re-verification found, and the fix

Re-running the verification at **121 frames** instead of 25 exposed a second defect, unrelated to the memory fix:

```
● Chain mode: 177 frames across 2 stages (tail 17)
✓ Saved: ./hdr-uat-stream.mp4 (177 frames, 704x448, 24 fps)
✓ Done — ltx-2.3-22b-distilled:fp8 in 377.4s

$ ls hdr-exr-stream | wc -l
0
```

121 frames exceeds the 97-frame `LTX2_DEFAULT_CLIP_FRAMES` routing default, so the CLI auto-chained. Both chain-stage builders pin `hdr_exr_dir: None`, so **zero EXR files were written while the run reported success**. A user asks for an HDR sequence, sees `✓ Saved`, and finds an empty directory — in a compositor, later.

That is the same failure shape this PR exists to prevent, so it now refuses the combination:

```console
$ mold run ... --frames 121 --hdr-exr-dir /tmp/out
error: --hdr-exr-dir cannot be combined with auto-chaining: 121 frames exceeds the
  97-frame clip this model auto-chains at, and EXR export writes one sequence per
  render rather than per clip. Render at most 97 frames, or drop --hdr-exr-dir and
  keep the tonemapped video.
```

No directory is created and the checkpoint never loads.

**Why refuse rather than support it now.** Threading the directory into stage requests would be *worse* than the error: each stage decodes from frame index 0, so stage 2 overwrites `frame_00000.exr` onward and the survivor is the last clip's tail, silently mislabelled as the whole render. Correct support needs frame numbering global across stages, which means accounting for the overlap frames `stitch_extend_frames` **drops**. Filed as #688 with that arithmetic spelled out.

Two tests: one pins the guard ahead of any other chain work, the other pins that both stage builders still drop the directory — so if chain stages ever *do* carry it, the guard gets revisited rather than left rejecting something that would work.

## On the verification itself

My original 25-frame run passed and would pass forever: 25 is below the auto-chain threshold, so HDR export and chaining never met. Both bugs found in this round — the 12 GB buffer and this — lived at a boundary a small clip cannot reach. The lesson I am taking is narrower than "test bigger": **test where a new feature meets features it does not know about**, because that seam is invisible to any single-subsystem happy path.

## Status

Rebased onto `main` (now carrying #682, #683). Full gate after the rebase: fmt, clippy `-D warnings`, feature-combo check, and `cargo test --workspace` **all clean with zero failures** — the first genuinely empty run this branch has had, now that #682 has fixed the `execution_plan` cluster.

Issue #685, opened 2026-08-03
An unclassified engine-shaping variable panics at generation time instead of failing CI


jamesbrink:
## The gap

`crates/mold-server/src/execution_plan.rs:587` fails loud on an engine-shaping variable it cannot classify:

```rust
// A new engine-shaping input must never silently collapse into an old
// equivalence class.
_ => panic!("unclassified frozen engine-shaping variable {name}"),
```

The intent is right — silently collapsing a new input into an existing equivalence class would corrupt execution-plan identity, which is far worse than a crash. **The timing is wrong.** Adding a name to `mold_inference::runtime_env::ENGINE_SHAPING_VARIABLES` without also adding a `RuntimeSemanticVariable` arm compiles cleanly, passes CI, ships, and then panics on the **first generation request that reaches planning**.

Nothing currently catches it:

- `runtime_env.rs:155` asserts a diagnostic variable is *not* in the list.
- `tests/execution_equivalence.rs:376` asserts the list's **length**.

Neither iterates the entries and classifies them, so the one invariant that matters is unenforced.

## Why it is worth closing

The two halves live in different crates — the list in `mold-inference`, the classifier in `mold-server` — so the natural edit (add the variable where the list is) leaves the other half untouched with no local signal. The failure then surfaces as a panic in the server's planning path, which reads like a scheduler bug rather than a missing match arm.

## Suggested fix

A test that walks the list through the classifier turns a production panic into a CI failure:

```rust
#[test]
fn every_engine_shaping_variable_has_a_semantic_class() {
    for name in mold_inference::runtime_env::ENGINE_SHAPING_VARIABLES {
        // Panics today for an unclassified name; the point is that it
        // panics here rather than on a user's first job.
        let _ = runtime_semantic_variable(name);
    }
}
```

Better still, make the classifier return `Option`/`Result` and have the *test* be the thing that rejects `None`, so the production path can report the misconfiguration without aborting a running server.

## Provenance

Found while adding `MOLD_LTX2_SPATIAL_TILE` for #673 — the variable was registered, compiled, and then panicked at generation time with `unclassified frozen engine-shaping variable MOLD_LTX2_SPATIAL_TILE`. Any future engine knob will hit the same trapdoor.

Issue #592, opened 2026-07-29
feat(ltx2): 4K / 48fps output path via upscaler composition


jdilley:
LTX-2.3 natively targets up to 4K output and 48/50 fps, reached via the spatial (x2 / x1.5) and temporal (x2) upscalers that mold already ships and exposes as request params (`Ltx2SpatialUpscale`, `Ltx2TemporalUpscale` in `crates/mold-core/src/types.rs`, as of 1503696d). But mold clamps requests to roughly 1 MP (`crates/mold-core/src/validation.rs`), so there is no supported path to 1080p/4K output even on hardware that can afford it.

**Scope**
- Relax the resolution clamp for LTX-2.x when reached via upscaler composition (base gen at <=1 MP, then spatial x1.5/x2 chained), with tiled decode where needed.
- A "target output" UX: user asks for 1080p/4K + 48 fps and mold composes base resolution + upscaler passes automatically, instead of users hand-computing stage shapes.
- Document VRAM expectations per rung; fail with actionable errors, not silent quality downgrades.
- Verify 9:16 portrait works across the allowed sizes and presets (validation is multiples-of-16, so 704x1216 should already pass — add tests + UI presets).

PR #694, opened 2026-08-04
feat(ltx2): tiled stage-2 refinement and spatial VAE decode tiling

### LINKED ISSUE #673: feat(ltx2): tiled stage-2 refinement for out-of-distribution resolutions
Prerequisite for the 4K half of #592, and for HDR at high resolution (#594).

## Why

LTX-2 checkpoints ship `positional_embedding_max_pos = [20, 2048, 2048]` and RoPE normalizes pixel positions by that span, so an axis past 2048px is outside the trained range — it renders, but out of distribution, with no error. #668 added a guard that rejects those shapes rather than producing them silently.

Upstream reaches 4K a different way: stage 2 runs over overlapping **latent tiles**, each denoised at a shape the model handles well, with positions renormalized per tile so every tile looks like a sequence starting at zero. Results are recombined with a separable trapezoidal window. `packages/ltx-pipelines/src/ltx_pipelines/hdr_ic_lora.py` (`:455-563`) is the only single-GPU reference; `ltx-core/modality_tiling.py` and `tiling.py` hold the machinery.

## Scope

- Tile arithmetic: interval splitting, blend window, tile enumeration. **Done** — `origin/tiling-primitives`, 10 tests, fixtures captured by running upstream's own `split_by_count` in its virtualenv.
- Wire it into `render_real_two_stage_av`: per-tile latent slice, per-tile positions via `build_video_positions` at the tile's pixel shape (which is what upstream's `normalize_positions=True` achieves), per-tile RNG seeded `seed + tile_idx`, accumulate the weighted results.
- Spatial VAE decode tiling, so the decode of a large frame does not undo the memory saving.

## A bug found while porting

`split_by_count` silently breaks the partition of unity once the stride drops below the overlap: tiles *i* and *i+2* then also overlap, three trapezoids stack, and the windows sum to as much as **1.16**. `blend` has no normalization pass, so the seam is brightened rather than erroring. Measured:

| split | window sum |
|---|---|
| `(21, 2, 8)` | 1.000 |
| `(21, 3, 8)` | up to 1.074 |
| `(21, 4, 8)` | up to 1.161 |

Upstream's `_clamp_tile_to_latent` only prevents the `ValueError`, not this. Solving `tile_size >= 2 * overlap` gives `overlap <= dim / (n + 1)`, which the port enforces. Upstream's shipped config is 2 tiles per axis, which can never triple-overlap, so they do not hit it — but any layout tuned for 4K would.

## Verification

The property that makes tiling *correct* is independent of the resolution that makes it *necessary*, so this can be verified without 4K hardware: assert a tiled stage-2 render matches an untiled one within a documented per-pixel tolerance at 512x512x25. True 4K needs 48-80GB per upstream's own table and is not verifiable on a 24GB card.

jamesbrink:
Closes #673. Prerequisite for the 4K half of #592.

## What this does

Wires the tile arithmetic from #674 into real execution. Stage 2 refines over overlapping **latent** tiles, each denoised at a shape the model handles well with its positions renormalized so every tile looks like a sequence starting at zero, then recombined through a separable trapezoidal window. Spatial VAE decode tiling composes with the existing temporal chunking, so decoding a large frame does not undo the memory saving.

`--spatial-tile <off|auto|PX[:OVERLAP]>` and `MOLD_LTX2_SPATIAL_TILE`. Auto-tiling only engages past the 2048 px RoPE span, which `mold_core::validation` does not admit today — so nothing currently renderable changes behaviour.

## A tiled stage 2 is video-only

It carries stage 1's audio through unrefined. A spatial tile says nothing about an audio track, and denoising the same track once per tile has no defensible recombination. Upstream does the same (`hdr_ic_lora.py:504-507`):

> Each tile calls `stage_2.run()` with a tile-sized `ModalitySpec` for **video only (audio is omitted entirely for HDR)**.

Half-disabling that branch makes the transformer reject the step outright, so `Stage2AudioInputs::video_only` turns shape, latents, noise, positions and context off as **one unit**. The exact rejection message is quoted in the doc comment so a future "simplification" greps straight to the explanation.

## The verification criterion in #673 was wrong, and I wrote it

The issue asked that a tiled render match an untiled one within a per-pixel tolerance. **That cannot hold by construction.** Each tile is denoised with its own RNG seeded `seed + tile_idx` — which is what makes tiles independent, and what upstream does — so a tiled render is a *different sample*, not a perturbation. Measured on the real pair (`ltx-2-19b-distilled:fp8`, seed 424303, 512x512x25, one with `--spatial-tile 256:64`):

```
max abs diff    234/255
mean abs diff   25.75/255
```

Any tolerance loose enough to pass that would also pass a badly broken implementation.

## What actually holds: seam-freedom

Mean horizontal gradient per column, so a seam appears as a localized spike. For 512 px with 256 px tiles and 64 px overlap the boundaries fall near x = 191/192, 255/256, 319/320:

| frame | x=192 | x=256 | x=320 | image mean | sharpest columns |
| --- | --- | --- | --- | --- | --- |
| f009 | 2.27 (z=+0.57) | 1.04 (z=−1.58) | 1.56 (z=−0.66) | 1.94 | 443–445, 494–495 |
| f015 | 2.28 (z=+0.48) | 1.05 (z=−1.65) | 1.90 (z=−0.19) | 2.00 | 445–448, 493 |

Every seam column sits **at or below** the image's own mean gradient, and the sharpest columns are image content far from any boundary. Two independent frames agree. The mild *softening* at seam centres is the trapezoidal blend averaging two estimates — the opposite of the hard edge a broken blend produces.

The two properties that a GPU cannot check stay unit tests, and they are where a partition-of-unity bug would hide:
- `tiled_refinement_reassembles_a_pointwise_denoiser_exactly` — the blend is exact (2x3 asymmetric layout, so a height/width transpose cannot survive).
- A one-tile plan short-circuits before the blend, so untiled output stays bit-identical.

`#![allow(dead_code)]` is gone from `tiling.rs`; every primitive now has a production caller.

## Note for reviewers

`ENGINE_SHAPING_VARIABLES` panics at **generation time**, not compile time, for an unclassified variable — adding the name is not enough, you must also add a `RuntimeSemanticVariable` arm. Filed separately as #685; not fixed here to keep this PR scoped.

## Gates

`cargo fmt` · `cargo clippy --workspace --all-targets -D warnings` · `cargo test --workspace` (zero failures) · `check:frontend` (architecture, studio/web/desktop tests, web + desktop builds) · `website` prettier 3.8.1 clean.


jamesbrink:
## Post-rebase GPU re-verification

The seam measurements in the description came from a pre-rebase binary. Merging lip-dub's `StageAudioConditioning` into the tiled call sites materially changed this code path — a conditioning decision hoisted out of the dispatch closure, a now-dead whole-frame start-latent removed, audio gated per pipeline — so I re-ran the pair on a binary built from the rebased tree rather than reporting numbers about different code.

Both renders took the real path (`render_native_video using real path pipeline=Distilled`), 512x512x25, seed 424303, `ltx-2-19b-distilled:fp8`, one with `--spatial-tile 256:64`:

| frame | mean abs diff | tiled grad mean | x=192 | x=256 | x=320 | sharpest cols |
| --- | --- | --- | --- | --- | --- | --- |
| f009 | 25.84/255 | 1.94 | z=+0.57 | z=−1.58 | z=−0.66 | 443–445, 495 |
| f015 | 25.72/255 | 2.00 | z=+0.48 | z=−1.65 | z=−0.19 | 446–448, 493 |

**Identical to the pre-rebase run to two decimal places**, including the sharpest-column positions. The merge preserved tiling behaviour exactly, which is the thing a re-verification is for — and worth more than my assurance that the changes "looked equivalent".

Every seam column still sits at or below the image's own mean gradient, with the sharpest gradients in image content far from any boundary. Two frames agree, as before.

The ~25.8 mean difference against untiled is unchanged and expected: per-tile RNG makes this a different sample, which is why the pixel-tolerance criterion originally specified in #673 cannot work. Wall time 398.3 s tiled vs 418.2 s untiled — tiling is not a regression at this size, though at 512x512 it is doing strictly more work for no benefit since auto-tiling only engages past the 2048 px RoPE span.

PR #740, opened 2026-08-06
fix(ci): lint GPU feature paths

### LINKED ISSUE #738: ci: clippy never runs with cuda/metal features, and mold-server has violations there
Found while landing #733 (for #598): **no CI gate has ever run clippy with a GPU feature enabled**, and `mold-server`'s CUDA-gated code has accumulated violations as a result.

## The hole

- `ci.yml` → `rust` runs `cargo clippy --workspace --all-targets -- -D warnings`, but with **default features** — no `cuda`, no `metal`.
- `ci.yml` → `cuda-check` / `metal-check` enable the backend but only run **`cargo check`**, which does not lint.

So every `#[cfg(feature = "cuda")]` block in the workspace is unlinted. #733's `flash-attn` job briefly ran `cargo clippy -p mold-ai --features flash-attn --all-targets -- -D warnings` (the feature implies `cuda`) and immediately turned `main` red on pre-existing violations; the step was narrowed to `cargo check` to unblock, so the debt is still here and still unguarded.

## What is actually there

Reproduce with:

```
cargo clippy -p mold-ai-server --features cuda --all-targets -- -D warnings
```

Four errors on `1ed31e23`:

| Location | Lint |
| --- | --- |
| `crates/mold-server/src/memory_preflight.rs:728` | `clippy::needless_return` |
| `crates/mold-server/src/memory_preflight.rs:821` | `clippy::needless_return` |
| `crates/mold-server/src/model_manager.rs:26` | `unused_imports` — `preflight_memory_guard_after_drop` |
| `crates/mold-server/src/memory_preflight.rs:762` | `dead_code` — `preflight_memory_guard_after_drop` never used |

The two `needless_return`s are cosmetic. **The other two may not be.** `preflight_memory_guard_after_drop` is imported by `model_manager.rs` and defined in `memory_preflight.rs`, and is called from neither. Given CLAUDE.md's invariant that a worker "may re-sample physical memory and reject the exact `predicted_vram_peak_bytes` if pressure changed… CUDA rechecks again after dropping replaced state", a memory guard named *after_drop* that nothing calls looks like a missing call site rather than something to delete. That needs someone who knows the scheduler's intent, which is why it was not fixed reflexively during a hotfix.

## Scope

- [ ] Determine whether `preflight_memory_guard_after_drop` should be called (and where) or removed — do not just silence the lint
- [ ] Fix the two `needless_return`s
- [ ] Audit `--features metal` the same way; it has had exactly as much lint coverage, which is none
- [ ] Add clippy-with-GPU-features to CI so this cannot recur. Cheapest option is upgrading `cuda-check` / `metal-check` from `cargo check` to `cargo clippy … -- -D warnings`; they already install the toolchains, so the marginal cost is the lint pass rather than a new job.
- [ ] Once green, consider restoring #733's `flash-attn` step to `cargo clippy` for the same reason

jamesbrink:
Closes #738.

## What changed

- run Clippy with warnings denied for CUDA and Metal forced-local feature paths, including all mold CLI targets
- restore Clippy for the flash-attn binary wiring now that CUDA-gated debt is clean
- remove the two CUDA-only needless returns
- box completed local batch responses to eliminate the backend-only large enum variant
- extend the CI routing contract so these GPU lint gates cannot silently regress to cargo check

## Validation

- full Metal forced-local Clippy with all targets
- 16 memory-preflight tests
- 2 local-batch tests
- CI routing contract
- cargo fmt and git diff checks

The CUDA command cannot run locally on macOS because nvcc is unavailable; this PR intentionally makes the Linux CUDA job the authoritative validation for that backend.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -289,6 +289,8 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy
       - name: Free runner disk space
         run: |
           sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc /usr/local/.ghcup
@@ -308,8 +310,8 @@ jobs:
         with:
           shared-key: workspace-cuda-sm89
           save-if: ${{ github.ref == 'refs/heads/main' }}
-      - name: Check CUDA forced-local and server paths
-        run: cargo check -p mold-ai --features cuda,preview,expand,tui,webp,mp4,mdns
+      - name: Clippy CUDA forced-local and server paths
+        run: cargo clippy -p mold-ai --features cuda,preview,expand,tui,webp,mp4,mdns --all-targets -- -D warnings
 
   # The `flash-attn` feature was decorative for months (issue #598) precisely
   # because nothing ever compiled it: the cfg-stripped fallback in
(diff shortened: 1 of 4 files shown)


jamesbrink:
CI follow-up: the FlashAttention PR run was taking the documented 79-minute cold path because its 587 MB cache had already been evicted from GitHub's repository-wide cache pool. This PR now excludes unrelated root-workflow edits from that expensive path filter. CUDA and Metal feature Clippy remain mandatory on the PR; full FlashAttention qualification still runs on every main push and on PRs that change the attention dispatcher or its Cargo feature wiring. The CI routing contract pins that policy.

PR #753, opened 2026-08-06
feat(core): add Wan 2.1/2.2 manifests, validation, and video authorities (#742)

jamesbrink:
Closes #742. Layer **1/5** of the Wan video stack ([tracking #741](https://github.com/utensils/mold/issues/741), GitHub stack #756, milestone [Wan Video (2.1/2.2)](https://github.com/utensils/mold/milestone/4)). Merge bottom-up: this PR first.

## What this adds

mold-core contracts for the Wan 2.1/2.2 video family — everything later layers build on, with no engine code:

- **Manifests**: `wan21-t2v-1.3b:bf16` (480p16 T2V bring-up model) and `wan22-ti2v-5b:fp16` (720p24 T2V+I2V), in the ComfyUI-repack weight layout the GGUF/LoRA ecosystem shares. Every file pins a SHA-256 taken from the HF API, and both manifests ship **hidden** so nothing unloadable is listed — the engine PR (#758) un-hides them.
- **Companions**: `wan-umt5` / `wan21-vae` / `wan22-vae` for future single-file catalog checkpoints. The 11.4 GB UMT5-XXL encoder deduplicates into `shared/wan/` between the companion and both primary manifests (`wan-umt5` declares `TextEncoder` shards like `flux2-te`, not a model-specific `Transformer`, precisely so storage collides).
- **Validation**: Wan's `4k+1` frame grid (`WAN_TEMPORAL_SCALE`), the flat 257-frame guard advertised on-grid, and per-checkpoint recommended buckets — the 480p-only 1.3B and the 704-grid TI2V-5B must not inherit each other's sizes, so `/api/models` resolves buckets per model with the family list only as the unknown-checkpoint fallback.
- **Video authorities**: Wan is registered everywhere the repo treats "video family" as data — `ExpandTask::for_family`/`for_conditioning` (T2V/I2V classification) **and its browser twin** `studio/lib/expandTask.ts` per the conditioning-aware-expansion invariant, `normalise_output_format` (MP4 default), and a format gate rejecting image formats.
- Upstream's tuned Chinese negative prompt as the manifest default, and a Wan prompt-expansion template.
- The frames-grid error message now derives its examples from the family's own step instead of hardcoding LTX's.

## Verification

- All SHAs/sizes verified against real HF file metadata; the manifests' component sets are exercised by round-trip tests (name resolution incl. legacy-dash forms, storage-path dedupe, VAE-generation separation).
[...]

Issue #741, opened 2026-08-06
Wan 2.1/2.2 video family: T2V + I2V (tracking)


jamesbrink:
Native Rust implementation of the **Wan 2.1 / 2.2** video family in mold: text-to-video **and** image-to-video, CUDA-first on 24 GB consumer GPUs (RTX 4090 primary target).

Benchmark experience: the HF space [`cinderholm/wan2-2-i2v-v3`](https://huggingface.co/spaces/cinderholm/wan2-2-i2v-v3) — fast, high-quality I2V. Reverse-engineering it shows the speed is not exotic: **Wan2.2 I2V-A14B with the lightx2v Lightning 4-step LoRA pre-fused** (checkpoint `thornmaze/WAMU_v3_WAN2.2_I2V_LIGHTNING`), **CFG off** (guidance 1.0, one forward/step), **UniPC flow-matching** (`flow_shift 3.0`, 6 steps default, range 4–8), fp8 W8A8 linears, 832×480-class output @ 16 fps (default 53 frames), RIFE fps interpolation as a post step. Ranked speed contributions: step distillation ~7×, CFG-off ~2×, fp8 ~2×, AoT compile 1.53× (mold is already AOT native code), small resolution. All of it is replicable in mold; only fp8 GEMM and flash-attn are CUDA-only.

## Model targets (phased)

| Phase | Model | Params | Geometry | Why |
|---|---|---|---|---|
| bring-up | `Wan2.1-T2V-1.3B` | 1.3B | 480p, 81f @ 16fps | fits trivially (~2.7 GB bf16); golden-parity iteration vehicle |
| first 720p | `Wan2.2-TI2V-5B` | 5B dense | 704×1280, **121f @ 24fps** | officially "24 GB (RTX 4090)" supported; one checkpoint does T2V **and** I2V |
| headline | `Wan2.2-T2V-A14B` + `Wan2.2-I2V-A14B` | 2×14B MoE | 480p/720p, 81f @ 16fps | the space experience; GGUF Q5/Q8 + Lightning 4-step, one expert resident at a time |
| deferred | `Wan2.1-I2V-14B` | 16.4B | 480p/720p | needs a bespoke open-CLIP ViT-H vision tower; Wan2.2 dropped CLIP entirely, so skip unless demanded |

## Architecture cheat-sheet

[...]

PR #754, opened 2026-08-06
feat(inference): add the UMT5-XXL encoder for the Wan family (#743)

### LINKED ISSUE #743: wan: UMT5-XXL text encoder (bf16 + GGUF, per-layer relative bias)
Part of the Wan video family (tracking: #741). Stack layer 2.

## Scope (`wan/02-umt5`)

New `crates/mold-inference/src/wan/text/umt5.rs` — self-contained UMT5-XXL encoder (do **not** modify the candle fork's `t5.rs`; precedent: `encoders/t5_gguf.rs` reimplements its stack in-crate).

Config: vocab **256384**, d_model 4096, d_ff 10240, 64 heads, d_kv 64, 24 encoder layers, relative buckets 32 / max_dist 128, gated **tanh-GELU** FFN, no attention scaling (T5 convention), eps 1e-6.

The one structural delta vs candle's `t5.rs`: **`shared_pos=False`** — every layer owns its `relative_attention_bias` and recomputes position bias per block (candle computes it in block 0 and threads it through).

- [ ] BF16 safetensors path (Comfy-Org repack `umt5_xxl_fp16.safetensors` / diffusers `text_encoder/` shards; state-dict prefix handling for `text_encoders.umt5xxl.transformer.…` full-checkpoint form)
- [ ] GGUF path via `city96/umt5-xxl-encoder-gguf` (llama.cpp t5 arch naming with per-layer `enc.blk.N.attn_rel_b` — `encoders/t5_gguf.rs` already handles that tensor class; token_embd force-dequantized like ComfyUI-GGUF does), wired into the quantized-encoder auto-fallback + `variant_resolution.rs`
- [ ] Tokenizer: `google/umt5-xxl` SentencePiece via shared_pool; EOS=1, PAD=0, 512-token truncation
- [ ] **Load-bearing output contract**: run with attention mask, then **trim each prompt's embeddings to true length and zero-pad to 512** — padded positions must enter the DiT as zeros, not as T5 pad-token outputs
- [ ] Drop/park/reload + CPU placement surface copied from `encoders/t5.rs` (`drop_weights`/`reload`/`park_to_cpu`); existing `T5_VRAM_THRESHOLD` applies as-is (UMT5-XXL is the same size class as T5-XXL)

## Acceptance
- Golden test: tokenization + embedding shape/trim-pad contract; GGUF-vs-BF16 cosine-similarity smoke on a fixed prompt
- clippy -D warnings, workspace tests green; CHANGELOG entry

Key facts: `tmp/wan-research/report-official-wan.md` §4, `report-candle-mold-gap.md` §B.1.

jamesbrink:
Closes #743. Layer **2/5** of the Wan video stack ([tracking #741](https://github.com/utensils/mold/issues/741), stack #756). Base: `wan/01-core` (#753).

## What this adds

The UMT5-XXL text encoder (5.7B params, vocab 256 384) as a self-contained stack in `wan/text/umt5.rs`.

**Why self-contained rather than candle's `t5.rs`**: UMT5's defining trait is `shared_pos=False` — *every* layer owns its relative attention bias and recomputes position bias from its own embedding, where T5 computes it once in block 0 and threads it through. candle's implementation hard-codes the shared layout, and per project policy the same-name candle patch branch is reserved for backend-only changes. Precedent: `encoders/t5_gguf.rs`.

- **Two weight sources, one module tree**: HF-named safetensors (the Comfy-Org `umt5_xxl_fp16.safetensors` repack — layout verified against its real header: 24 blocks × 10 tensors, per-block bias `[32, 64]`) and city96-named GGUF (`enc.blk.{i}.attn_rel_b` at every block).
- **Wan's load-bearing output contract**: encode with the attention mask, then zero every position at/beyond each prompt's true length — the DiT must see embeddings of the zero vector for padding, not T5 outputs for pad tokens. Pinned by a mask-isolation test (junk tokens in the padded tail must not perturb real positions).
- **Prompt canonicalization** matching upstream's `clean='whitespace'` (double HTML-unescape, whitespace collapse, strip; ftfy's mojibake pass deliberately skipped and documented).
- **EOS-preserving truncation**: overlength prompts cap content at 511 and keep `</s>` in the final slot, matching HF truncation semantics.
- Refuses plain T5/mT5 checkpoints with an error naming the offending block.
- Drop/reload retention surface mirroring `encoders/t5.rs` so the transformer gets the VRAM during denoise.

## Verification

- Codex-reviewed pre-submit; 2 findings fixed here (EOS truncation, canonicalization).
- Relative-bucket math pinned against hand-computed HF values; 6 unit tests.
- Proven on real weights downstream: both #758 and #759 UATs encode through this module (including the CPU-parked F32 path fixed in #759).


jamesbrink:
Closes #743. Layer 2 of the Wan stack (#741, stack #756). Codex-reviewed pre-submit: 2 findings fixed (EOS-preserving truncation, upstream clean='whitespace' prompt canonicalization with documented ftfy deviation).

jamesbrink:
@greptileai please review

Issue #830, opened 2026-08-07
perf(h3): add block streaming and exact memory admission


jamesbrink:
Part of the MiniMax H3 native audio-video milestone (tracking: #814). Add H3-specific block streaming, phase-aware memory lifecycle, and authoritative host-RAM/VRAM admission within Mold's existing one-device-per-stage Scheduler V2 contract.

## Execution strategy

- [x] Add `minimax_h3/offload.rs` and an H3 admission authority parallel to the specialized LTX-2 path.
- [x] After conditioning, release Qwen GPU residency before denoising. Stream transformer blocks from admitted host memory to the one leased GPU with bounded resident depth, prefetch, and cancellation.
- [x] Phase the visual VAE encode/decode and keep the small FP32 AudioVAE resident when practical to avoid pathological repeated streaming.
- [x] Freeze task partition, checkpoint/config hashes, Qwen truncation, quantization profile, attention backend, resident-block count, prefetch depth, VAE tiling, reference shapes, frame grid, and audio policy at admission.
- [x] Fingerprint and charge every transformer shard, Qwen shard, processor asset, VAE, AudioVAE, quantization sidecar, and pruned AdaLN table independently.

## Exact estimates

- [x] Derive packed text/condition/audio/video row counts after preprocessing and before placement.
- [x] Price full-attention and FFN workspaces, Qwen vision/language activations, condition VAE encodes, target video latent/noise, stereo audio latents, decoder tiling, AAC mux staging, resident blocks, dequantization buffers, and prefetch overlap.
- [x] Reserve all streamed bytes in the host ledger while preserving Mold's safety floor (`max(15% total RAM, 8 GiB)`).
- [x] Advertise low estimate confidence only for genuinely pending artifacts; recompute from landed bytes at queue admission.
- [x] Reject full 134.1 GiB stacks on hosts that cannot preserve the floor. Document that the approximate 75 GB offload claim is marginal on 96 GB systems and that 128 GB host RAM is the realistic initial recommendation for the curated low-memory tier.

## Scheduler boundary

- [x] Lease exactly one GPU for each H3 generation under Scheduler V2. Batch siblings may use different GPUs, but one render cannot.
- [x] Do not borrow an unleased GPU for Qwen or VAE phases and do not claim tensor/Ulysses/pipeline parallel support.
- [ ] If future two-GPU support is pursued, require a separate design for atomic device bundles, per-device fingerprints, transport/collectives, cancellation, deadlock-free acquisition, and UI reporting.

[...]

Issue #827, opened 2026-08-07
test(h3): qualify hardware, performance, release, and docs


jamesbrink:
Part of the MiniMax H3 native audio-video milestone (tracking: #814). Close the family only with real-hardware proof, cross-framework parity, operational baselines, regressions, and honest user documentation. Governance approval is complete in #831.

### Numerical and semantic parity

- [ ] Run every #832 layer against the pinned official full-precision implementation and archive revision/component hashes plus tolerances.
- [ ] Prove Comfy pruned INT8/NVFP4 output against the full path with spatial/temporal perceptual metrics, audio spectral/loudness/channel metrics, and exact A/V timing.
- [x] Lock T2VA prompt/noise ordering, each FL2VA endpoint signature, reference ordering/rotary clock, scheduler sign/shifts/grid, and 124/243/362-frame geometry.

### Real CUDA UAT

- [ ] T2VA at 124 and 362 frames, 1344x768, synchronized 32 kHz stereo MP4.
- [ ] FL2VA first-only, last-only, and both endpoints, including mismatched source aspects.
- [ ] Ref2VA image-only, video-with-soundtrack, image+standalone-audio, mixed image/video/audio order, swapped-order comparison, and every count/duration failure.
- [ ] Full/resident high-memory tier where hardware is available and the documented 24/32 GB streamed tier on the declared consumer qualification GPU.
- [ ] Cancellation during Qwen load/encode, visual/audio reference encode, block streaming, denoise, visual decode, audio decode, mux, and gallery persistence.
- [ ] Missing/partial/corrupt components, unavailable attention backend, host-memory pressure, VRAM pressure, instance change, network interruption, and runtime-policy withdrawal.

## Benchmarks

- Record exact GPU/driver/backend, CPU, host RAM, storage, component hashes/bytes/dtypes, attention kernel, resident blocks, prefetch, shape/reference mix, grid points/evaluations, cold/warm setup, Qwen, per-step denoise, VAE, AudioVAE, mux, total wall time, peak VRAM, and peak host RAM.
- Compare official Diffusers, ComfyUI curated workflow, and Mold on the same feasible machine/shape when license authority allows; distinguish source claims from independently reproduced data.
- Publish results for 960x544 and 1344x768 at 124 frames, plus 362-frame feasibility. Never imply hosted Regenerate-2K parity.
- Add performance budgets/regression thresholds only after a stable baseline exists; do not encode aspirational upstream numbers as CI truth.

## Documentation and release proof

[...]

PR #924, opened 2026-08-09
feat(h3): bind private runtime qualification evidence

jamesbrink:
Summary:
- add a development-only producer that authenticates the private artifact qualification and owner-only external capture evidence, then emits deterministic review candidates for all thirteen runtime memory bounds
- bind each candidate to the exact campaign source, complete local server dependency closure, locked dependency graph, canonical build features/environment, measured server ELF, CUDA route, attention authority, and retained evidence identities
- normalize only the reviewed record-allowlist value region to avoid an executable-hash fixed point; every other runtime/build input change invalidates the campaign
- share the 128 KiB producer/activation limit, reject unsafe paths and mutable evidence, and keep both producer and private reader excluded from release artifacts
- include every runtime-identity build script/support input in the Docker dependency-cache layer and enforce that build-context contract
- document that the producer authenticates the retained ELF while external profiler/log process binding remains a live-campaign evidence and review responsibility

Validation:
- runtime-code identity and build-axis tests: 5/5
- producer adversarial tests: 6/6
- reviewed activation tests: 5/5
- conditioner producer contracts: 21/21
- full Linux CUDA distribution contract: 36/36 parser tests plus Docker and release assertions
- focused Rust 1.93 Clippy with warnings denied and ordinary server check
- private-UAT release contract, JSON schema, Rust formatting, diff check, prohibited-term scan, and exact conflict-free merge tree against merged main

This does not add a reviewed runtime record, activate the private server path, or claim measured GPU qualification. It creates the fail-closed producer and reproducible authority boundary needed for that separate campaign.

Advances #827 and #830.


jamesbrink:
## Review pass — two P1s found and fixed

Codex xhigh review plus an independent compliance audit. Both findings are fixed in `d7563cd`.

**1. `cargo publish -p mold-ai-server` would have failed, after the release tag was already pushed.**

`crates/mold-server/build.rs` reached into `../mold-inference/build_support/` for the canonical campaign feature validation. A published `.crate` carries only files beneath its own crate directory, so the packaged build script cannot compile. Proven directly — copying the packaged `build.rs` into an isolated directory and compiling it gives:

```
error: couldn't read `../mold-inference/build_support/h3_runtime_code_identity.rs`: No such file or directory
```

`cargo package -p mold-ai-server --list` confirms nothing from `../mold-inference/` is staged. The helper now lives in `crates/mold-server/build_support/h3_server_features.rs` (which packages, and compiles standalone), and the unused `sha2` build-dependency is gone with it.

Nothing in CI could have caught this: `scripts/tests/crates-publish-contract.sh` mocks `cargo publish` entirely, and no other job packages a crate. That contract now fails when any published crate's build script loads source from outside its own directory — written first, watched fail on the old code, then fixed.

Splitting the file means two copies of `CANONICAL_H3_SERVER_FEATURES`, so the release contract pins them together. Mutation-verified: changing one entry yields `FAIL: the canonical private H3 server feature set drifted between mold-inference and mold-server`.

**2. The runtime-code identity could not distinguish two CUDA toolkits.**

`CUDA_HOME` and the host compiler variables were captured as strings — they record *where* the toolchain is, not *what* it is. A toolkit upgraded in place under the same prefix leaves every captured string, `rustc -vV`, and every source input identical while producing a different executable, so an unchanged identity could authorize runtime code that was never measured.

The identity now binds what `nvcc` and the host C compiler report about themselves, and the build script watches both binaries so replacing either invalidates a cached identity (a cached identity would otherwise outlive the toolchain, since an in-place upgrade changes no tracked input). A CUDA campaign build whose `nvcc` cannot be resolved or run fails rather than producing an identity that cannot tell two toolkits apart. Non-CUDA builds — including CI's `cargo clippy -p mold-ai-inference --features dev-bins,h3-private-uat` — take an explicit absent branch and need no toolkit.

## Compliance re-check

Unchanged by this branch, verified independently of the PR description: `REVIEWED_RUNTIME_QUALIFICATION_RECORD_SHA256` is still `&[]`, `crates/mold-core/src/model_policy.rs` is untouched by the diff, and `scripts/verify-h3-release-exclusion.sh` rejects the new producer marker against a synthetic binary. No manifest, URL, catalog row, engine arm, weight-derived fixture, or activation override is added.

## Gates run locally

`crates-publish-contract.sh`, `minimax-h3-private-uat-release-contract.sh`, `cuda-distribution-contract.sh` (which caught the missing `build_support/` COPY in the Docker dependency-cache layer), `cargo fmt --all --check`, `cargo clippy -p mold-ai-inference --tests -D warnings`, and `cargo test -p mold-ai-inference --test h3_runtime_code_identity` (7/7, two of them new).

jamesbrink:
## Second review pass — the toolchain binding named the wrong compiler

A codex xhigh review of `d7563cd` found that binding *a* toolchain identity is not enough if it can describe a different compiler than the one cargo actually used. Verified against the vendored source, not taken on faith:

`cudaforge` resolves nvcc as `NVCC` → `PATH` → `CUDA_HOME` → `CUDA_PATH` (`cudaforge-0.1.6/src/toolkit.rs:108-135`) and hands the host compiler to `nvcc -ccbin` from `NVCC_CCBIN` (`cudaforge-0.1.6/src/builder.rs:393`). My resolver ignored `NVCC` entirely, preferred the CUDA-root variables over `PATH`, and asked `CC`/`cc` about the host compiler — so two builds using genuinely different native compilers could land on the same runtime identity, which is worse than binding nothing.

Fixed in `db29d74`. Precedence now mirrors cudaforge exactly, the host compiler comes from `NVCC_CCBIN` before `CC`, and the resolved absolute paths join the identity next to the version banners (two toolkits can print the same banner from different prefixes). The precedence is a pure function pinned by a test covering each rung, including the case that caught me — an `NVCC` override pointing at a nonexistent path must fall through to `PATH`, not to `CUDA_HOME`.

## One gap documented rather than claimed closed

Watching the compiler binaries reruns only Mold's own build scripts. The CUDA objects come from `candle-kernels` and `candle-flash-attn`, whose cached outputs Mold cannot invalidate, so a toolchain replaced under an existing `target/` produces an executable linking previous-compiler objects while the identity reports the new one — and nothing in the build can detect it.

That is a campaign procedure requirement, not a code-enforced invariant, so `docs/qualification/minimax-h3.md` now says so plainly and requires a clean target directory for the measured build. Flagging it here because it is the kind of residual that should be an explicit reviewer decision rather than something absorbed silently into a fix commit.

jamesbrink:
## Third pass — the mirror was incomplete in two ways (`df67350`)

Severity dropped to P2, and both are the same shape: my mirror was *stricter* than cudaforge, so each would have failed a campaign build whose kernels cudaforge compiles without complaint — the opposite of the intended fail-closed behaviour.

- **The resolver stopped after cudaforge's first four branches.** Its fifth sweeps standard prefixes (`cudaforge-0.1.6/src/toolkit.rs:8-15,138-163`), so a `/usr/local/cuda` install absent from `PATH` resolved for cargo and not for the identity. The `PATH` lookup also accepted any file of the right name where cudaforge uses `which`; a non-executable `nvcc` is not the program cargo ran.
- **`NVCC_CCBIN` is normally a directory.** `nvcc --compiler-bindir` documents the directory form, and executing it as a program aborted the build with a confusing error. A directory value now resolves to the runnable compiler inside it, and keeps the configured value when there is none so the failure still names what was set.

Both verified against the vendored cudaforge source rather than inferred, and covered by tests: each rung of the search order, the `/usr/local/cuda`-not-on-`PATH` case, directory-valued ccbin, plain-executable passthrough, and a directory holding only a non-executable `cc`.

## Stopping the review loop here

Three passes, severity P1 → P1 → P2, with each round's findings verified before acting on them. The remaining surface is the depth of the identity design itself, which is this PR's own subject and your call as owner rather than something to keep grinding on in review. Flagging that explicitly instead of either continuing indefinitely or implying the design is now exhaustively verified.

The one gap that is documented rather than closed remains the dependency-crate CUDA object cache described in the previous comment.

PR #925, opened 2026-08-09
feat(h3): add exact BF16 Qwen layer-50 capture producer

jamesbrink:
## Summary

- add an authorization-bound official Qwen layer-50 producer for text-only and representative multimodal cases
- add a dense CUDA Mold adapter that executes exactly the first 50 BF16 language layers without a quantized or approximate fallback
- bind raw inputs, tensor coordinates, model/runtime identities, owner-only external evidence, and release exclusion
- keep the real 66.7 GB capture explicitly unclaimed until a qualifying sufficient-memory CUDA route is available

## Validation

- 17 Qwen producer contract tests
- 21 conditioner producer contract tests
- private GPU conformance contract
- private-UAT release contract
- Rust 1.93 clippy with warnings denied
- exact conflict-free merge tree against the parent head

## Stack

Ready stacked on #924. Merge the parent first; this branch will then be rebased onto the resulting main head before merge.


jamesbrink:
## Review pass — two P1s, both blocking the documented workflow (`cf7c543`)

Codex xhigh review. Neither was visible to the contract tests as they stood, because both fail before the capture reaches a GPU.

**The shard geometry check could never balance.** `TEXT_ENCODER_PAYLOAD_BYTES` is the index's summed *tensor payload*, while the loop summed whole-file sizes — which also include each shard's eight-byte little-endian length prefix and its JSON header. Every official checkpoint was rejected, for both capture roles. Payload now comes from each shard's own header, which also validates that the declared header length is in range.

**`--role mold` could never stage its source.** It archives the whole repository, and Mold tracks `AGENTS.md -> CLAUDE.md`, so `git archive` always emits a symbolic link that the extractor rejected as a non-regular member. Confirmed directly against this repo:

```
$ git archive --format=tar HEAD | tar tvf - | grep '^l'
lrwxrwxrwx root/root         0 AGENTS.md -> CLAUDE.md
```

Links are now materialized as regular copies of an in-snapshot target, resolved and fenced inside the snapshot, after the regular files land. An escaping or dangling link still fails, non-regular members that are not links still fail, and the staged tree stays link-free so `snapshot_files` is unchanged.

Verified end-to-end against a real `git archive HEAD` rather than only the synthetic fixture: 1740 files staged, zero links, `snapshot_files` accepts 1934 entries. Before the fix that raised `CaptureFailure`.

Four new contract tests, written first and watched fail: symlink materialization, escaping-link rejection, non-link irregular-member rejection, and the payload derivation with its truncated-header case. Codex re-reviewed the fix commit and returned no findings.

## Rebased

Rebased onto the current `feat/h3-runtime-qualification-record` head, which now carries three review fixes of its own (including a release-blocking `cargo publish` failure). Full contract suite green on the rebased tree. A `CHANGELOG.md` entry is added — the repo requires one per feature and this stack had none.

Issue #783, opened 2026-08-07
feat(wan): chain/sequence wave — capability arm, last-frame handoff, Studio/Discord eligibility, extend


jamesbrink:
## Motivation
Split from #769 ("chain/sequence wave"). Wan is capped at one clip: `capability_for_family` has no wan arm, `supports_sequence` is advertised `false`, the CLI rejects `--frames` beyond the 257 cap instead of auto-chaining, sequence UI on web/desktop/iPhone filters wan models out, and Discord `/sequence` is hard-gated to `family == "ltx2"`. LTX-2 shipped its sequence wave months after its engine — this is the expected next layer, but wan needs a different handoff design because it has **no latent motion tail**: the natural smooth handoff is last-frame image conditioning, which requires an I2V-capable checkpoint, so eligibility must ride the same per-checkpoint conditioning classification as the source-image capability issue, never the family.

## Evidence
- `crates/mold-inference/src/chain/capability.rs:42-59` — only `ltx2` (TemporalHandoff) and `ltx-video` (IndependentClips); wan `chain: false` in `batch.rs`.
- `crates/mold-core/src/catalog.rs:61-63,142-143` — `chain_capable_family` is `matches!(family, "ltx2" | "ltx-video")`, driving `supports_sequence`/`supports_extend`.
- `crates/mold-cli/src/commands/chain.rs:99` — `is_chain_capable = family == Some("ltx2")`; over-cap wan requests are `Rejected`, never chained.
- `studio/lib/chainRouting.ts:135-141` — `CHAIN_CAPABLE_FAMILIES = {ltx2, ltx-video}`, `FAMILIES_WITH_CONTEXT_HANDOFF = {ltx2}`; `studio/lib/sequence.ts:72-75` family fallback knows only ltx.
- `crates/mold-discord/src/commands/sequence.rs:22,29,63,271` — four `"ltx2"` literal gates.
- Handoff raw material: `crates/mold-inference/src/wan/conditioning.rs:250` (`WanTi2vInpaint`) and `:331` (`per_token_timesteps`) already implement per-token-timestep latent pinning (built for frame 0, extensible to leading frames); A14B I2V conditions via the 36-channel mask+latent concat; A14B/1.3B T2V have no conditioning channel at all.

## Proposed work
[...]

PR #936, opened 2026-08-10
feat(wan): multi-clip sequences with a per-checkpoint handoff

jamesbrink:
Stacked on #935. Closes the core of #783.

Wan was capped at one clip: no arm in `chain/capability.rs`, `supports_sequence` advertised false, the CLI **rejected** an over-cap request instead of chaining it, the Studio pickers filtered wan out, and Discord's `/sequence` was hard-gated to four `"ltx2"` string literals.

## The handoff is a checkpoint property, not a family one

Wan has **no latent motion tail**. Its smooth handoff is last-frame *image* conditioning, which only an image-conditioned checkpoint accepts — so eligibility rides the checkpoint. `wan_carryover` is a total function of the `source_image` contract `wan::pipeline::source_image_capability` already derives from the checkpoint's own headers, so the two cannot drift:

| Checkpoint | Seam |
| --- | --- |
| `wan22-ti2v-5b:*` (`Optional`, latent inpaint) | continues |
| `wan22-i2v-a14b:*` (`Required`, 36-channel concat) | continues |
| `wan21-t2v-*`, `wan22-t2v-a14b:*` (`Unsupported`) | independent clips |

A text-to-video checkpoint is still sequence-capable — it concatenates, the honest LTX-Video precedent, and `transitionLabel` already degrades "Smooth" to "Join" at tail zero so no fake "Continue motion" is offered. An unclassified checkpoint is "unknown", never an assumed handoff. `capability_for_family("wan")` reports that conservative floor because the family view cannot know which checkpoint is selected; `capability_for_model` is for callers holding one.

## Two defects I found in my own earlier revisions of this branch

Both were caught by asking whether the engine could honour what I had advertised, and both are fixed in `f86eaf03`:

1. **`WanEngine` never implemented `ChainStageRenderer`.** `as_chain_renderer()` returned the trait's default `None`, so every wan sequence would have failed at worker dispatch — an advertised capability the worker could not execute. The renderer now drives the ordinary generation path through a lossless APNG round-trip; that decoder moved to `chain::encode` so it and the LTX-Video renderer share one copy instead of keeping two chances to disagree about channel order.
2. **I sized the seam from LTX-2.** I had carried over the 17-frame motion tail — the pixel window LTX-2's VAE turns into three latent slots of carryover. Wan's continuation re-renders exactly the one frame it was seeded with, so the seam is **1**. 17 would have discarded sixteen good frames at every boundary while looking entirely plausible. The routing test now asserts the seam directly, because at 53-frame clips a tail of 0 and a tail of 1 need the same number of stages — the stage-count comparison it replaced would have passed whatever the tail was.

## Also

- Chain-limits recommended clip frames snapped to a hardcoded `8n+1`, so wan's own shipped defaults came back off its `4k+1` grid and into a 422 from the validator that owns the real rule. Step and offset now come from the family.
- Auto-chain routing default is a VRAM envelope, not a ceiling: 53 for the two-expert A14B, 121 for the single-expert 5B, both on-grid. `--clip-frames` still wins, clamped to the real 257 cap.
- Discord's gates are now `mold_core::catalog::chain_capable_family` — the same predicate `/api/models` uses — so bot and server cannot disagree.
- `sequenceFrameOptions` takes the family and steps by 4 for wan.

## UAT (RTX 4090, real render)

```
● Chain mode: 145 frames across 3 stages (tail 1)
✓ Done — wan22-ti2v-5b:q8 in 141.2s (145 frames, seed: 11)
```

704x384, three auto-chained stages. The boat, gutter, railing, and foliage persist across both seams (47→49 and 96→120) — the scene holds, no drift to a different subject. A first attempt at `--steps 6` rendered noise in **stage 1, before any handoff was involved**; that was me ignoring the non-distilled model's 20-step recipe, not the chain, and I verified it by checking frame 0 rather than assuming.

## Not in this PR

Item 6 (wan `extend_video`) and the `/api/generate/chain/validate` + chain-TOML coverage from item 3 are not here. The capability, routing, renderer, grid, Studio, and Discord layers are, and they are consistent with each other and with the engine.

## Verification

`cargo test` on mold-ai-inference (1739), mold-ai-core, mold-ai-catalog, mold-ai, mold-ai-discord; `cargo test -p mold-ai-server --lib -- chain_limits variant_dependencies` (43); `bun run check:frontend` rc=0; `cargo clippy` on all five touched crates with `-D warnings`; `cargo fmt --all -- --check`; website prettier. The full `cargo test --workspace` sweep is left to CI — it deadlocks locally in `mold-server`'s `routes_test` where the blocked tests all take the shared `env_lock()`, which I did not establish reproduces on `main`.


jamesbrink:
Pushed `92b57653`: **#783 item 6 (extend) is now in this PR**, so the "Not in this PR" note in the description no longer covers it.

Wan rejected `extend_video` outright. It is the same seam as a sequence boundary with the carryover coming from a file: the source's final frame becomes the continuation's conditioning, the continuation re-renders exactly that frame, and the stitch drops it. The overlap is therefore always **1** on wan — a larger value is refused with a message naming why, rather than silently trimming good frames, because LTX-2's multi-frame overlap is a latent motion tail wan does not have. The stitch reuses LTX-2's `stitch_extend_frames` rather than a second chance to be off by one.

The overlap grid comes from the family now: wan's VAE compresses time by 4 where LTX-2's compresses by 8, so the hardcoded `8k+1` had rejected every valid wan overlap. `supports_extend` is advertised per checkpoint from the same `source_image` contract the seam reads, and the server re-derives it from its header probe so a config path override cannot leave a stale manifest guess promising a continuation a text-to-video checkpoint cannot accept. Resolution and fps are locked to the source and a mismatch is refused rather than rescaled.

UAT on the 4090: 48-frame source + 49-frame continuation − 1 overlap = **96 frames in 53 s**, boat and gutter continuous across the join.

Still not in this PR: the `POST /api/generate/chain/validate` and chain-TOML coverage from item 3.

jamesbrink:
Pushed `2cc5bb8c`: **#783 item 3's remainder is in**, so nothing on the issue is now outside this PR.

The chain preflight checked the per-clip cap but never the frame grid, so a wan stage off `4k+1` survived to the engine's request validator and failed mid-chain instead of being named at submission. The grid comes from the family now and the error names the stage. `motion_tail_frames` is normalized for wan the way it already was for ltx-video — 17 becomes 1, or 0 on a text-to-video checkpoint — with the contract probed from the resolved checkpoint's headers first, exactly as `/api/models` and generation admission do.

The `mold.chain.v1` script path needed no code change and is now documented and verified end to end:

```
$ mold chain validate wan-shot.toml
OK — 2 stages, 97 frames estimated
```

49 + 49 − 1 = 97 confirms the one-frame seam through the script path too.

**All seven items of #783 are now in this PR** — capability arm, grid arithmetic, CLI/server auto-chaining and validation, the three Studio surfaces, Discord, extend, and docs — each with a real render or validation behind it rather than an assertion.

Issue #802, opened 2026-08-07
perf(wan): engine efficiency follow-ups — per-token timestep tables, LoRA-branch staging, expert-swap prefetch, VAE decode batching/tiling


jamesbrink:
## Motivation
Four measured-or-analyzed engine inefficiencies found during the wan review, each measurement-first and independently landable. Grouped here so the milestone tracks them without four micro-issues; split any item out when picked up. Cross-references: item 2 interacts with the `CastBoundary` audit in the step-time P1 issue; items 1-2 return VRAM that feeds the 81f-reclaim envelope.

## 1. TI2V-5B per-token timestep tables are O(tokens) but hold exactly two values
`per_token_timesteps` builds `vec![t; total]` with the first `tokens_per_frame` slots zeroed — two distinct values by construction (`crates/mold-inference/src/wan/conditioning.rs:331-360`). Yet at 1280x704x121f (27,280 tokens) each forward materializes `timestep_proj` as `[1, 27280, 6, 3072]` F32 ≈ 2.0 GB (`crates/mold-inference/src/wan/model/transformer.rs:1050-1053`), and **every one of the 30 blocks** materializes another such table in `modulation()` via `broadcast_add` (`:664-679`), x2 per step under CFG (5B default guidance 5.0). The sinusoid is also computed host-side in f64 per forward after a GPU→CPU `to_vec1` round trip (`:1031-1037`, cost acknowledged at `:729-734`). Fix: embed the 2 unique values (`[2, 6, dim]` after the MLP) + `index_select` per token; hoist the embedding so the CFG pair shares one per step; kill the device→host→device round trip for the scalar path too (`pipeline.rs:502-518`). Pin equality with golden fixtures.

## 2. GGUF LoRA branch stages ~1.2 GB F32 and adds ~1,200 kernel launches per forward
[...]

PR #939, opened 2026-08-10
perf(wan): step-skip caching and A14B partner-expert prefetch

### LINKED ISSUE #801: perf(wan): step-skip caching (FBCache/TeaCache class) for the 20-step quality tiers
## Motivation
The A14B `:q8` quality tier runs 20 steps x two forwards (CFG 3.5) — ~10-20 minutes/clip at the measured ~30 s/step — and the 5B runs 20 steps at 121f. The Wan ecosystem's standard mitigation is residual caching across near-duplicate steps: TeaCache ~2x on Wan 2.1, MagCache 2.68x with better fidelity (has a Wan 2.2 port), FBCache/Comfy-WaveSpeed 1.5-3x with a tunable `residual_diff_threshold`. Crucially none of these help the 4-step Lightning tier — no redundancy to skip (field-confirmed, kijai/ComfyUI-WanVideoWrapper#811) — so this is **quality-tier-only** leverage and must be gated as such. Composes with the fp8 tier (per-step cost) rather than competing with it.

## Evidence
- wan-reports/report-speed-quant.md §5 — TeaCache (ali-vilab/TeaCache/TeaCache4Wan2.1), MagCache (Zehong-Ma/MagCache), Comfy-WaveSpeed FBCache; "compatibility with 4-step Lightning: effectively none".
- `crates/mold-core/src/manifest.rs:5147-5148` — quality tier: 20 steps, cfg 3.5; 5B is 20 steps (`:4855-4860`).
- `crates/mold-inference/src/wan/pipeline.rs:492-533` — the denoise loop is one clean per-timestep call into the transformer; a first-block-residual cache slots in at one boundary.

## Proposed work
1. FBCache-style first-block residual reuse (smallest surface, no offline calibration): compare block-0 output distance step-over-step; below threshold, skip the remaining blocks and reuse the cached residual. MagCache's calibrated per-model ratios are the fallback mechanism if FBCache quality disappoints.
2. Opt-in knob (`MOLD_WAN_STEP_CACHE=off|auto|<threshold>`), default off until quality is characterized; hard-disable when `steps <= 8` or a distill adapter is active; cond/uncond cached independently; cache resets at the expert-swap boundary; skipped-step count disclosed in progress output.
3. Determinism: threshold-based skipping is deterministic for a fixed request, but document that changing the threshold changes outputs; keep the default path untouched so the family's determinism contract in `batch.rs` stays honest.
4. UAT: fixed-seed side-by-sides at several thresholds on `:q8` A14B and 5B; record speedup vs artifact onset.

## Acceptance criteria
- ≥1.5x measured quality-tier wall-clock improvement at a threshold with no visible artifacting on the UAT prompts; numbers recorded.
- Fast tier and low-step requests provably unaffected (guard + test); cache-disabled runs bit-identical to today.
- Knob documented per the Tier-1 convention (configuration.md, SKILL.md, CHANGELOG).

jamesbrink:
Closes #801.

FBCache-style first-block residual reuse for Wan's non-distilled tiers. Block 0 runs every step; when its residual moves less than the relative-L1 threshold, blocks 1..N are skipped and their previous contribution is replayed.

## Measured (RTX 4090, `wan22-t2v-a14b:q8`, 33f @ 832×480, 20 steps, fixed seed)

| `MOLD_WAN_STEP_CACHE` | Time | vs off |
| --- | --- | --- |
| off | 605.6 s | — |
| `0.05` (first `auto` guess) | 620.5 s | **slower** |
| `0.10` | 327.4 s | **1.85×** |

No visible artifacting at 0.10. Be precise about what it buys: a cached run is a **different sample of comparable quality**, not the same frames faster — skipping steps changes the trajectory. That's stated in the docs rather than implied away.

## The `auto` default is measured, not guessed

I first shipped 0.05 reasoning that `auto` should be conservative. It **never fired once in 20 steps** and came out slightly slower than uncached, having paid the distance check for nothing. A conservative default that never engages is a knob that lies about being on, so `auto` is now 0.10. Comfy-WaveSpeed independently ships 0.09 for Wan.

## Guards

Neither is overridable, because both describe configurations where reuse cannot help and can only cost quality:

- a shipped distill adapter — the 4-step Lightning tiers have no near-duplicate steps (field-confirmed, kijai/ComfyUI-WanVideoWrapper#811);
- a schedule under 12 steps.

Both **refuse with a message** rather than silently doing nothing. Verified on `:q5`: `step cache ignored: a distilled adapter is active, which leaves no near-duplicate steps`.

Cond and uncond keep independent caches so their trajectories are never crossed, and the cache resets at the A14B expert swap where the network itself changes.

## Determinism

With the cache off the engine is bit-identical to before. The test pinning that also proves an always-accept threshold really does change the result — otherwise the identity half would pass vacuously on a cache that never fires, which is exactly the failure mode the 0.05 measurement caught.

`MOLD_WAN_STEP_CACHE` is registered in both `ENGINE_SHAPING_VARIABLES` and `RuntimeSemanticVariable`, so a cached run can never share an execution-equivalence class or a learned-timing bucket with an uncached one.

## Verification

- `cargo test -p mold-ai-inference --lib` — 1741 passed
- `cargo test -p mold-ai-server --lib execution_plan::` — 57 passed
- clippy `-D warnings`, `cargo fmt --all --check`, website prettier — all clean


jamesbrink:
Added #802 item 3 (A14B expert-swap prefetch) to this branch — same engine, same measurement setup, so it reviews together.

**The swap was a cold read.** The "page cache still holds it" rationale in `experts.rs` covers the *outgoing* expert; the incoming 10.8–15.4 GB file hasn't been read this run and, on a 32–64 GB host, has been evicted behind UMT5 (11.4 GB) plus the resident expert. mold now warms it on a background thread while the first expert denoises.

**Host I/O only** — bytes are read sequentially and discarded, so no VRAM is touched and the max-of-pair invariant is unchanged. A test asserts exactly that: one resident expert throughout, and the warm joined at the swap rather than left competing with the load it exists to accelerate.

Measured on an RTX 4090 with the pair evicted from the page cache (`wan22-t2v-a14b:q5`, 33f @ 832×480):

| Run | Swap load | Total |
| --- | --- | --- |
| cold, prefetch off (1) | 8.7 s | 95.0 s |
| cold, prefetch off (2) | 22.2 s | 131.1 s |
| prefetch on (1) | **5.6 s** | 86.9 s |
| prefetch on (2) | **5.6 s** | 90.9 s |

The cold number varies with what survived in cache; the warm number doesn't, because what's left is the GGUF parse and the upload. `MOLD_WAN_PREFETCH=0` disables it and is registered as engine-shaping so the two never share a learned-timing bucket.

Item 4 (VAE decode batching / spatial tiling) is **not** in this branch — it touches the causal-conv3d frame loop and the tiled-VAE registry, and belongs on its own.

Issue #776, opened 2026-08-07
perf(wan): reclaim 81-frame A14B clips on 24 GB — attention, F32 activation scoping, partial block offload


jamesbrink:
## Motivation
Split from #769 ("reclaim 81f on 24GB"). A14B's trained clip length is 81 frames; mold ships 53f (fast) / 33f (quality) defaults because the Q5 pair peaks at 23,975 MiB at 53f/832x480 and 81f OOMs on an RTX 4090 (`crates/mold-core/src/manifest.rs:5171-5183`). The ecosystem's standard levers all exist in reach: FlashAttention is already eligible at these call sites, the per-block F32 residual round trips are scopeable, and partial block streaming (Kijai's `blocks_to_swap`, ComfyUI's budgeted lowvram loader) is the community's 81f@720p answer — mold owns the pattern in `flux/offload.rs`, and the Wan DiT's 40 uniform blocks are the friendliest shape for it.

## Evidence — where the VRAM goes at 81f
- **Attention transients.** Wan self/cross attention routes through `crate::attention::attention` (`crates/mold-inference/src/wan/model/transformer.rs:582`); without `flash-attn` the CUDA path chunks queries at 512 (`crates/mold-inference/src/attention.rs:206-218`) but each chunk still materializes `[B*H, 512, N]` score + softmax tensors. At 81f/480p N = 21x30x52 = 32,760 tokens; with 40 heads that is ~1.3 GB per transient at BF16, more on the F32 softmax path. Flash is *already eligible* — the quantized path runs BF16 activations on GPU (`transformer.rs:120-144`) and `flash_is_eligible` accepts head_dim 128 (`attention.rs:284-305`). This is distinct from #736 (shipping flash-attn artifacts) and #734 (dense FLUX/Z-Image call sites): Wan already reaches the dispatcher; this issue is measurement and qualification.
- **Per-block F32 residual round trips.** `WanBlock::forward` casts the full hidden state to F32 for both gated residual adds and back (`transformer.rs:699-722`): several `[1, 32760, 5120]` F32 transients (~671 MB each) per block, x40 blocks.
- **LoRA branch staging** on the fast tier holds ~1.2 GB of F32 adapter weights beside the 10.8 GB expert (`crates/mold-inference/src/wan/lora.rs:325-331`) — BF16 staging returns ~0.6 GB (see the engine-efficiency follow-ups issue).
[...]

PR #954, opened 2026-08-11
feat(wan): lossless host/device moves for a quantized block

jamesbrink:
Groundwork for [#776](https://github.com/utensils/mold/issues/776) item 3 — partial block offload for the Wan DiT. This lands the mechanism and its correctness proof; the driver follows.

## Why this is needed

From the calibration re-fit in #949: an 81-frame A14B q5 render extrapolates to **~25,461 MiB against a 24,564 MiB card**. One of the 40 blocks is ~270 MiB at Q5, so parking a handful in host RAM covers the ~900 MiB shortfall with margin — a long way short of the stream-everything mode `flux/offload.rs` runs, and its 3-5x cost.

## Two things had to be settled first, and both are settled here

**1. The move must not go through a dtype round trip.**

The obvious spelling — dequantize to F32, move, re-quantize — is lossy, and silently so: `QTensor::quantize_onto` takes an F32 source, so a parked block would come back different from the one the checkpoint shipped, and the render would depend on how many blocks happened to be parked. candle exposes both halves needed to avoid that:

- `QTensor::data()` returns the raw quantized bytes, and **its CUDA arm is implemented** (`candle-core/src/quantized/mod.rs:249`) — so this works from a resident block, not only from a file.
- `ggml_file::qtensor_from_ggml(dtype, bytes, dims, device)` rebuilds a `QTensor` on **any** device from those bytes.

Composed, a park/unpark cycle is a memcpy each way. `parking_a_block_is_byte_identical` asserts raw storage equality rather than closeness, deliberately: a tolerance check is exactly what a lossy round trip would pass.

**2. It has to work at the tensor level, not on the block struct.**

`WanLinear` is `Arc<dyn Module + Send + Sync>`. A block cannot be asked to change device — there is no `to_device` on a trait object and no way to write one generically. De-erasing it would mean a `to_device` arm for the plain, GGUF, fp8, `CastBoundary`, and LoRA-branch variants plus every construction site in the family.

Moving the *tensors* instead lets the existing `WanBlock::load` path rebuild the block. That is the path initial load already uses, so **the quantized source and the LoRA parallel branch compose here exactly as they do today** rather than through a second, parallel implementation that could drift.

I originally posted the var-builder rebuild as the cheaper option on #776, then withdrew it: `mold_candle::quantized::VarBuilder::from_gguf` uploads every tensor eagerly (`quantized.rs:43-52`), so holding a var-builder to rebuild from would pin the whole expert on the GPU and defeat the purpose. `from_qtensors` being public is what makes the tensor-level route work.

## What is here

- `qtensor_to_device` — the lossless move, with an `Arc`-clone fast path when the tensor is already on the target device.
- `ParkedBlock` — one block's weight set, with `to_device`, `size_in_bytes`, and `into_tensors` (shaped for `VarBuilder::from_qtensors`).
- `WanBlockParking::park` — selects a block's tensors by `blocks.{i}.` prefix; returns `None` for an absent index so a caller discovers the real block count rather than trusting a config field against an unfamiliar export.
- `blocks_to_park` — the sizing arithmetic, kept separate from *which* blocks to park: the count is arithmetic worth pinning, the choice is a scheduling decision for the driver.

## No consumer yet

Deliberate, and consistent with this family's existing convention — `#[allow(dead_code)] pub mod wan`, commented "built and tested here ahead of the layers that consume them". The driver, its resident-set policy, prefetch, the `block_offload` registry flip, the scheduler disposition, and 81f UAT follow in the next change. Splitting it this way keeps the correctness argument above reviewable on its own rather than buried in a streaming loop.

## Verification

- `cargo test -p mold-ai-inference --lib wan::` — 199 passed (5 new)
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo clippy -p mold-ai-inference --lib -- -D warnings` — clean (checked separately; the lib-only target is what hid a dead-code warning on #936)
- `cargo fmt --all -- --check` — clean

Note: codex peer review could not run on this one — the local Codex CLI is returning 401 `token_expired` and rejects every model. Verification above is mine.


jamesbrink:
Peer review (subagents) found a real defect in this PR's own test, now fixed in 0dea1573.

**`parking_a_block_is_byte_identical` proved nothing.** `to_device(&Device::Cpu)` on CPU tensors hits the same-device short circuit and returns the input `Arc`, so the test compared a tensor with itself — it would have passed no matter how broken the round trip was, and with no GPU in CI nothing else covered it. That was the module's entire correctness claim.

Fixed by splitting `rebuild_on` out as the no-short-circuit path and pointing the test at it. Production callers keep the short circuit, and `a_same_device_move_returns_the_same_allocation` now pins both halves on purpose: the fast path returns the same allocation (`Arc::ptr_eq`), the rebuild path does not, and the rebuilt bytes still match.

Two gaps also closed:

- `every_shipped_quantization_survives_the_round_trip` — Q4_K, Q5_K, Q8_0, F16, F32, since `qtensor_from_ggml` is per-dtype and the other tests only exercised one.
- `a_parked_block_reloads_through_the_var_builder` — pins the contract the driver depends on: `into_tensors()` feeds `VarBuilder::from_qtensors` and resolves at `.pp("blocks").pp(i)`. This is why `park` keeps full checkpoint names rather than stripping the prefix; `VarBuilder::path` joins `pp` segments and looks the result up verbatim, so getting it backwards fails at runtime with `cannot find tensor`, not at compile time. It also asserts a sibling block's weights are not reachable from the set.

Also dropped the hand-written `same_device` for candle's own, which it duplicated arm for arm.

**One gap documented rather than closed**, because it belongs to the driver: `WanTransformer` does not retain the checkpoint tensor map — `from_gguf_with_loras` drops the var builder after construction — so `park` cannot be called post-construction yet. The module notes why the map cannot simply be kept alive: `from_gguf` uploads every tensor eagerly, so holding it would pin the whole expert on the GPU and defeat the purpose.

8 tests pass; `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --all -- --check` clean.

jamesbrink:
Correcting the caveat in b6728160: **the allocator hypothesis is refuted — parking genuinely frees VRAM.**

That commit recorded 8 blocks parked moving peak only 21,354 -> 21,066 MiB and said the mechanism was unproven until explained. Ran the decisive test — same shape, 30 blocks parked:

| Parked blocks | Peak | Freed |
| ---: | ---: | ---: |
| 0 | 21,354 MiB | — |
| 8 | 21,066 MiB | 288 MiB |
| 30 | **15,082 MiB** | **6,272 MiB** |

30 blocks is ~8.1 GB of Q5 weights and 6.27 GB came back, so the freed memory does return to the pool and is reusable. candle's caching allocator is not swallowing it.

The 8-block figure was the anomaly, not the rule: relief is non-linear at small parking counts, consistent with the allocator having already grown its pool to the unparked high-water mark and simply not needing to reuse the freed blocks at that point. Practical consequence for the policy — parking a *few* blocks is not a reliable way to claw back a few hundred MiB, so a shortfall should be covered with margin rather than by the exact arithmetic minimum.

Correctness is unaffected either way: 53f at 8 blocks parked rendered a coherent frame, which is what proves the lossless byte round trip on real CUDA.

Still outstanding on this issue and NOT in this PR:
1. **Admission relief.** `memory_preflight` still refuses 81f at ~25,461 MiB before the engine ever loads, so the offload cannot yet rescue the shape it exists for. The relief belongs on the weight term, which this PR does not touch.
2. **81f UAT**, which is gated on (1).
3. Item 4 (raising the A14B defaults), gated on (2).

I stopped short of changing shared admission code rather than rush it; flagging rather than burying.

PR #1068, opened 2026-08-14
fix(h3): placement previews return promptly instead of hanging in Planning

jamesbrink:
## Summary

Tracks down and fixes the H3 "hitting Generate goes into Planning… and never returns" report on plato, plus the fleet-wide `mold-0.10.0` Nix store-path version mismatch.

**Root causes found on plato (openat tracing + symbolized perf):**
1. The placement preview's private admission SHA-256-verified all ~42 GB of H3 artifacts from scratch, once per candidate GPU (~170 GB of hashing per preview on 4× L40S).
2. `OpenedSupportArtifact::revalidate()` re-hashed the full Qwen support set on every call — invoked ~60×/s during admission.
3. `authenticate_full_sha256` fully re-read the Qwen NVFP4 text-encoder artifact per authority open.
4. `stage_weight` copied ~5.8 GB of VAE weights into a fresh tempdir per open, per candidate device.
5. Fresh public-H3 deployments never created `MOLD_HOME/h3-runtime/staging`, so the VAE load plan failed on every request.
6. Preview media redaction sent present-but-empty FL2VA endpoints that admission rejected outright.
7. No client abort ceiling, so all of the above rendered as an unbounded Planning spinner.
8. `flake.nix` hardcoded `version = "0.10.0"` in two places, so deployed 0.21.0 binaries ran from `/nix/store/…-mold-0.10.0` paths.

**Fixes:** process-lifetime caches keyed on complete metadata identity (dev/inode/size/mode/owner/mtime/ctime) for artifact digests, opened-checkpoint hashes, Qwen full authentication, and staged VAE pairs — each with per-path single-flight so concurrent cold fills share one pass; one-time-full-rehash support revalidation matching the model artifact's own precedent; cheap fail-closed policy pre-checks before bulk hashing; owner-only staging-dir bootstrap; contract-validated placeholder substitution for redacted preview endpoints (gated behind request validation); a 15-minute client abort ceiling with a self-disarming Safari 15 fallback; and `workspaceVersion` everywhere in `flake.nix`.

## Measured on plato (4× L40S, `--features cuda,h3`)

| Preview | Before | After |
| --- | --- | --- |
| Warm | ~9m48 (often perceived as never returning) | **2.4 s** |
| Cold (first request after server start) | indefinite hang / many minutes per retry | 2m19, resumable via digest caches |

Both return `outcome: "planned"` end-to-end through the real `POST /api/generate/placement-preview`.

## Test plan
- `cargo test -p mold-ai-inference --features h3-private-uat --lib minimax_h3` — 206 passed (macOS)
- `cargo test -p mold-ai-candle --features h3-private-uat --lib minimax_h3` — 188 passed (macOS)
- Same suites on plato under `--features cuda,h3` — all passing except the 6 pre-existing `private_server::tests::reviewed_record_*` cases that reproduce identically on unmodified main (h3 never runs in CI; regenerating the reviewed record needs compliance campaign materials)
- `studio` vitest `generationPlacement` — 23 passed
- `cargo fmt` + clippy `-D warnings` (h3-private-uat, all targets) clean
- Codex review addressed: single-flight added for all cold cache fills (its P2), placeholder allocation gated behind contract validation (earlier P1)

https://claude.ai/code/session_01RJpcRagserkopZeCakLnKm


jamesbrink:
Correction follow-up: #1082 removes the Unix owner/write-mode admission policy introduced here. Ordinary model files and model roots may be shared or group/other writable; H3 retains pinned-content, regular-file/symlink, canonical-containment, and descriptor-identity checks. Credentials and retained private evidence remain owner-only.

Issue #1055, opened 2026-08-14
fix(ltx2): restore image-to-video conditioning and preprocessing parity


jamesbrink:
## Summary

LTX-2 text-to-video is producing good results in Mold, while image-to-video frequently produces one of two failure modes:

- little or no meaningful motion (often a static image or simple zoom), or
- motion that looks like a localized filter/deformation and does not make physical or semantic sense for the source photo.

This strongly isolates the problem to the image-conditioning path rather than the shared LTX-2 transformer, scheduler, text encoder, or decoder. A source comparison against the official Lightricks implementation found two high-confidence parity defects plus a related preprocessing mismatch cluster.

**Primary hypothesis:** Mold currently (1) skips the training-distribution H.264 preprocessing required by official LTX-2 I2V and (2) applies source conditioning twice before the first transformer evaluation while calculating the timestep as if it were applied once. Together, those defects can explain both under-motion and incoherent local motion.

## User-visible reproduction

1. Select an LTX-2 model that works well for text-to-video.
2. Generate T2V with a motion-explicit prompt and confirm coherent motion.
3. Use a realistic photo as the first-frame/source image with the same model and a motion-explicit prompt.
4. Leave source strength near the current default (`0.75`) and otherwise use ordinary settings.
5. Observe that many inputs remain almost static or acquire implausible warping/filter-like motion.

The issue is easiest to see with realistic phone photos, portrait subjects, and source/output aspect-ratio mismatches. Exact seed/model/source/prompt fixtures should be captured while implementing this issue so the comparison remains reproducible.

## Finding 1: required H.264/YUV420 conditioning preprocessing is missing

The official LTX-2 pipeline does not feed a pristine still image directly to the VAE. It deliberately:

1. decodes with orientation and color-profile handling,
2. encodes the still as a one-frame H.264/YUV420 video at a model-version-specific CRF,
3. decodes that compressed frame,
4. aspect-preserving resizes and center-crops it to the requested canvas, and
5. normalizes and VAE-encodes the result.

This is not a cosmetic export step. Upstream documents the CRF round-trip as matching the conditioning distribution used during training:

- [Official image preprocessing and H.264 round-trip](https://github.com/Lightricks/LTX-2/blob/fd4ded7f2d88d3da713abcdd4ad41ecc4a9314ca/packages/ltx-pipelines/src/ltx_pipelines/utils/media_io/decode.py#L386-L435)
[...]

PR #1076, opened 2026-08-15
ltx2 i2v parity 3 conditioning

jamesbrink:
The pre-loop freeze blend re-applied the clean target to start latents
that apply_video_token_replacements had already soft-blended, so an I2V
source at strength s entered the first transformer pass as
(2s-s²)·C + (1-s)²·N while the per-token timestep still claimed (1-s)·σ —
a latent/timestep mismatch that suppressed global motion (93.75% source
at the 0.75 default instead of 75%). Upstream constructs the noisy state
with a single lerp (noisers.py:32-33 @ fd4ded7).

Extract the init into initialize_video_stage_latents: the freeze blend
now runs only for an explicit caller-supplied clean tensor (retake's
temporal window; lip-dub's frozen stage-2 audio on the audio mirror),
where it is the single lerp that installs the source into mask=0 regions
of pure-noise start latents. For conditioning-derived cleans it was
either the double-count (soft replacements) or an exact identity (hard
replacements, appended keyframes/references/chain anchor — latents and
clean hold the same tokens), so T2V and every hard path are bit-identical.

Witnesses: initialized latents hold exactly s·C+(1-s)·N for
s ∈ {0, 0.25, 0.75, 1}, masked timesteps match the actual noise fraction,
appended tokens keep their clean values, the retake shape still installs
frozen regions, and T2V passes through untouched.

Part of #1055 (stacked PR 3/4, on #1072). This is the core motion fix.

Scope note vs the issue text: the issue asks to remove the pre-loop blend; implementation shows that blend is load-bearing for retake (it installs the source video into mask=0 regions of pure-noise start latents) and is exactly upstream's noiser lerp — the true duplicate was applying it on top of the already-soft-blended replacements. The fix therefore keeps the blend only for explicit clean overrides and leaves start latents un-re-blended otherwise, which is provably bit-identical for T2V, keyframes, chain/extend carryover, IC-LoRA/lip-dub references, and retake — only soft-strength I2V sources change (by design). The appended-soft-token timestep observation (chain anchor s=0.4 initializes un-noised under a 0.6σ timestep) is a pre-existing mold-specific chain behavior, deliberately unchanged here; follow-up issue to come.

Seeded before/after render evidence + upstream parity matrix will be attached before merge (UAT running on the RTX 4090).

### DIFF
--- a/crates/mold-inference/src/ltx2/runtime.rs
+++ b/crates/mold-inference/src/ltx2/runtime.rs
@@ -2631,6 +2631,67 @@ fn maybe_apply_temporal_upsampler(
     Ok(upsampled)
 }
 
+/// The video-modality state entering the denoise loop: conditioned start
+/// latents, widened positions, the per-step clean target, and the denoise
+/// mask.
+struct VideoStageInit {
+    latents: Tensor,
+    positions: Tensor,
+    clean: Tensor,
+    denoise_mask: Tensor,
+}
+
+/// Build the initial conditioned video latents exactly once (#1055).
+///
+/// Upstream constructs the noisy state with a single lerp
+/// (`noisers.py:32-33` @ fd4ded7): a conditioned token starts as
+/// `s·C + (1-s)·N` and its per-token timestep says `(1-s)·σ` — the two
+/// must agree. `apply_stage_video_conditioning` already performs that
+/// blend for soft replacements (the i2v source path), so re-blending the
+/// result toward the clean target — which this function's pre-fix caller
+/// did for every conditioned render — double-counted the source:
+/// `(2s-s²)·C + (1-s)²·N` under a timestep still claiming `(1-s)` noise.
+/// That latent/timestep mismatch suppressed I2V motion at soft strengths.
+///
+/// The freeze blend itself is still upstream's lerp and remains
+/// load-bearing when the caller supplies an *explicit* clean tensor over
+/// pure-noise start latents (retake's temporal window; lip-dub's frozen
+/// audio mirrors this on the audio side): there the blend is what installs
+/// the source into the frozen (mask = 0) regions, and it runs exactly
+/// once. For conditioning-derived cleans it is removed rather than kept:
+/// hard replacements and appended tokens already equal their clean values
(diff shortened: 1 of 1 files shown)


jamesbrink:
## Seeded before/after UAT (RTX 4090, `ltx-2-19b-distilled:fp8`, seed 42, 49f @ 24fps, 768×512)

Full report + videos/strips retained at `tmp/uat-1055/` in the working tree (`REPORT.md`, `run-matrix.sh` has the exact commands).

**T2V regression gate: bit-identical.** Same seeded T2V request on the pre-stack baseline (`1bb4467b`) and this branch produces byte-identical MP4s (framemd5 `85a0964e…`). The conditioning fix provably does not touch T2V.

**I2V opening-frame fidelity** — PSNR of frame 0 against the aspect-correct fitted source (photoreal 768×1024 portrait onto the 768×512 canvas):

| strength | baseline | fixed |
|---:|---:|---:|
| 0.50 | 11.69 dB | 27.07 dB |
| 0.75 | 11.79 dB | 28.31 dB |
| 1.00 | 11.96 dB | 28.83 dB |

The baseline's ~12 dB is geometry, not noise: `resize_exact` squashed the portrait source to the landscape canvas — its frame 0 scores **31.7 dB against a deliberately squashed reference** while the fixed build scores 12 dB against that same distorted reference. The old path faithfully pinned a distorted image; the new path pins an undistorted center-crop (#1055 Finding 3), with the single-blend init (Finding 2) and the CRF-33-equivalent round-trip (Finding 1, PR #1072) upstream of it.

**Visual review** (first/mid/last strips): baseline shows the squashed opening frame, identity drift, and background progressively replaced by filter-like rigging/vine clutter; the fixed build keeps the source composition anchored (boat cabin, railing, face geometry) with purposeful prompted motion (raises the mug, drinks) at every strength, and s=1.0 pins the opening frame while motion still develops. Residual mid-clip background artifacts at low strength match upstream's own I2V reports (Lightricks/LTX-2 #11/#255) — this restores parity, not perfection.

Codex high-reasoning review of this PR's diff: no findings.

Issue #814, opened 2026-08-07
MiniMax H3 native synchronized audio-video (tracking)


jamesbrink:
Native Rust/Candle implementation of **MiniMax H3 Base**, with the official implementation as the semantic/numerical oracle and ComfyUI's pruned single-file path as the practical deployment target.

> [!IMPORTANT]
> **Governance complete:** #831 authorizes H3 use in every territory and across local, remote-client, shared-server, hosted, generated-output, distribution, and redistribution paths. The README and H3 user guide are the complete required user-facing license/NOTICE/attribution/disclosure/AUP delivery; no H3-specific product control is required. Technical availability remains limited to implemented and qualified model, task, request, hardware, and artifact routes.

## Product target

| Mode | Inputs | Output |
| --- | --- | --- |
| T2VA | prompt | 5-15 second video with synchronized stereo audio |
| FL2VA | prompt plus zero, one, or two first/last-frame images | conditioned video with synchronized audio |
| Ref2VA | ordered heterogeneous references: up to 9 images, 3 videos, 3 audios, and 12 total files within documented duration limits | reference-conditioned video with synchronized audio |

Initial support is CUDA-only, 24 fps, 768p-class, dimensions divisible by 32, native batch size 1, and the official 5-15 second duration/frame grid. Mold's Batch N remains ordered singleton siblings. Audio is part of the diffusion trajectory and cannot be treated as an optional post-process.

## Architecture and load-bearing parity points

- 33.1B dense, 50-layer single-stream omni transformer: hidden size 5376, 56 attention heads, head dimension 128, FFN 14336, video latent width 24, audio latent width 32, and patch size `(1,2,2)`.
- Qwen3-VL-32B conditioner, truncated at the documented layer and consumed without its final normalization; H3 tokenizer/special tokens and prompt processor are part of the contract.
- f16t4d24 visual VAE: causal encoder plus ViT3D decoder, spatial compression 16, 17-frame chunk geometry, per-channel latent normalization, and exact condition-image preprocessing/posterior sampling.
- DAC/BigVGAN-lineage AudioVAE: stereo 32 kHz, 40 latent Hz, 800 samples per latent step.
- Packed `[text | ordered condition blocks | target audio | target video]` sequence, modality tags, 3D multimodal RoPE, full attention, and modality-specific AdaLN.
[...]

PR #1169, opened 2026-08-19
perf(h3): correct the denoise device ledger — 24 GiB-class admission

jamesbrink:
## What

Re-aligns the MiniMax H3 compact FL2VA device ledger from L40S-era policy constants to correct, enforced arithmetic. Four changes, one theme — the denoise phase now charges what is actually live:

1. **Workspaces max, not sum** — attention and FFN transients never coexist; both ledger authorities share `denoise_transient_workspace_device_bytes`. (~−10.1 GiB)
2. **Enforced liveness** — all three block forwards explicitly drop attention-phase tensors before the MLP allocates, so the ledger's assumption is code, not hope (found by adversarial codex review of the first cut).
3. **VAEs parked across denoise** — both VAEs drop after condition encoding and are reconstructed from a second authenticated authority (opened at admission, bound to the canonical validation identity) before decode; new `VaesParked`/`VaesReloaded` phase states and load-drop policy variant; visual decode charges the reconstruction workspace. (~−5.8 GB)
4. **Explicit hidden-activation baseline** — the transformer's live rows×5376 BF16 activations were previously charged nowhere (absorbed by workspace-cap slack); now an explicit 4-tensor term with the liveness derivation in the doc comment. (+1.6 GiB, honesty tax)

Net: predicted denoise peak at the qualified 1344×768/124f envelope drops ~33.6 GB → **~19.4 GB**, and the advisory capability tier drops 40 GiB → **24 GiB** ("Compact 24 GiB VRAM"). The 40 GiB figure traced to an uncommitted L40S review; the L40S was a dev box, never a target (#814 comment).

## What this does NOT do

- Host-RAM admission still predicts ~61.8 GB increment and will refuse a 64 GB host — the mmap-aware host accounting fix is the next PR. A 4090 box passes the VRAM check after this PR but not yet the host check.
- No instrumented render yet (none has ever run in-repo); the tier is corrected analytic arithmetic, and #827's measurement campaign on the 4090 will re-derive it from `observed_bytes`.

## Testing

- TDD both ledger changes (test-side reference flipped first; validator tests red → green).
- New `parked_vaes_are_charged_around_denoise_but_never_across_it` pins the parking property; the phase-ledger structural test asserts via drop sentinels that the VAEs are absent across all three denoise forwards and reconstructed before decode.
- `cargo test -p mold-ai-inference --features h3-private-uat --lib`: 2065 passed / 0 failed. `-p mold-ai-candle minimax_h3`: 190 passed. `-p mold-ai-server h3_private_bridge`: 11 passed. fmt + clippy (default and h3-private-uat) clean.
- Two `mold-ai-server` route tests fail locally from known env leakage (this machine has the real H3 weights installed; tests assert absence) — pre-existing, fails identically without this diff, CI is clean.


jamesbrink:
Codex adversarial review (local) surfaced a real P1 on the first cut: shadowed hidden-sized tensors (~408 MiB each at the qualified envelope) survived into `mlp.forward`, and the transformer's live hidden activations were charged nowhere — the old attention+FFN sum absorbed them silently. Remediation, in this PR:

1. `0f7c90a9` — all three block forwards (compact INT8, BF16, token refiner) now scope the attention-normalized input, drop the attention projection after its gated residual, and drop the MLP input after use, so the code enforces the liveness the ledger assumes.
2. Follow-up commit (pending, same PR): an explicit denoise hidden-activation baseline term — K x total_packed_rows x hidden_size x BF16 with K derived from the now-enforced worst-case simultaneous count — added to the evidence builder, validator, and test reference, keeping max(attention, ffn) honest instead of relying on workspace-cap slack.

Net effect stays ~8+ GiB below the old sum while becoming correct rather than accidentally conservative. Do not merge until (2) lands.

jamesbrink:
Item (2) landed: `9c2a3c8d` adds `denoise_hidden_activation_device_bytes` (K=4 × total_packed_rows × 5376 × BF16, derivation in the doc comment) to builder, validator, and test reference. The PR body reflects the final shape. Merge hold from the previous comment is lifted.

PR #1171, opened 2026-08-19
perf(h3): per-phase, anonymous-bytes-only host-RAM admission

jamesbrink:
Stacked on #1169 (device ledger). This corrects the host side — the last analytic admission blocker for a 24 GB / 64 GB (RTX 4090-class) machine.

## What

`predicted_host_increment_bytes` was a flat sum of ~21 terms: every artifact's full file bytes (42.48 GB of mmap'd/streamed safetensors), the whole Qwen host workspace (dropped after encode), VAE IO, staging, and mux buffers — charged as if simultaneous. Predicted ≥68.9 GiB; refused every 64 GB host.

Now it mirrors the device design: eleven per-phase host sums on the same phase boundaries, take the max, and count only genuinely anonymous demand:

- **Not charged anywhere**: artifact file bytes (the Qwen loader reads one tensor at a time into a bounded Vec, the transformer streams one packed block with a hard refusal of a second, the VAEs mmap — each decision commented at the term with the loader `file:line`) and the VAE's file-backed mapping — `MemAvailable`, this prediction's own input, already counts those pages free (#1108 class).
- **Still charged**: the Qwen packed CPU parameters (20.26 GB of real anon Vecs) during its phases, the single live transformer block, layout/schedule/noise buffers, media bounds.
- **Two undercharges fixed the other direction**: fixed-transformer load staging was double-counted as residency but missing its real transient shape, and the NVFP4 loader's largest-tensor double-buffer (1.56 GB) was measured but never budgeted.

Eleven new `*_phase_host_bytes` budget fields are hashed into the budget identity and relationally validated per phase, so builder and validator can no longer desync on a single phase silently — the property the device side already had.

Also: `H3_CURATED_HOST_RAM_RECOMMENDATION_BYTES` 128 GiB → 64 GiB (arithmetic in the doc comment; 32 GiB honestly does not fit: 24.5 GiB peak + 8 GiB floor).

## Numbers (real envelope, CPU-Qwen route)

| phase | GiB |
|---|---|
| qwen_encode (peak) | 24.46 |
| qwen_transfer | 23.01 |
| mux | 2.15 |
| old flat sum | ≥68.94 |

64 GB host: needs 34.06 GiB available incl. floor — fits. 32 GB host: needs 32.46 of 32.00 — genuinely refused.

## Testing

TDD per-phase (test reference first); new `host_demand_is_a_per_phase_max_of_anonymous_bytes_only` pins the max discipline, exact peak/denoise term composition, and a validator differential proving file-backed bytes move nothing. 13/13 `h3_factory`, 2065 `h3-private-uat --lib` (one known unrelated flake), fmt/clippy clean, workspace check clean. Three `mold-ai-server` route tests fail locally from pre-existing weights-on-disk env leakage — verified identical on the base commit.

## Known residual (deliberate)

`max_streamed_block_host_overlap_bytes` misses one `from_raw_buffer` copy (~2 GiB below the peak phase, cannot move the prediction) — same bug class, noted for the instrumented-run follow-up (#827).


jamesbrink:
Codex adversarial review fixes landed in `ac56fc06`: (1) `qwen_host_load_staging_bytes` is now derived — pinned `== 2 × maximum_tensor_staging_bytes` from the released loader facts at the same seam that pins parameter residency, carried on the frozen authority identity, and added to the relational-mutation test (passing `1` is rejected); (2) retained opened-component metadata (Qwen raw header 231,408 B, transformer header ≤8 MiB cap, VAE config Vecs 4,282 B) is now charged in exactly the phases its authority lives. Real-envelope peak effectively unchanged (~24.46 GiB + ≤8.23 MB); the 64 GiB tier stands.

PR #1178, opened 2026-08-20
feat(h3): wire the Turbo LoRA tiers through the inference runtime

jamesbrink:
Third PR of the H3 Turbo series (#1172 parser, #1174 forward). This makes a Turbo tier actually renderable: 8 transformer evaluations instead of 20 (9-point terminal-inclusive grid), or 4 with the 768p tier.

## What

- **Sampler**: `H3SamplerKind::ComfyEuler` — the existing pinned comfy-simple sigma grid with the euler integrator (Turbo's reference workflows use euler; grids pinned to exact f32 bits at 9@shift12, 5@shift12, 5@shift6, plus the Turbo README's published NFE=4 decimal oracle). The 768p 4-step tier's shift-6 schedule threads through `new_for_sampler_with_video_shift`; a non-default shift is refused for the integrators whose baked audio constants assume shift 12.
- **Factory authority**: `H3FactoryTurboAdapterAuthority` on the quantization authority (adapter identity/digest/tier/sampler/steps, hashed; schema v5→v6; the previously `..`-tolerant identity match is now an exhaustive destructure so a new field can never be silently unhashed). The runtime's hardcoded `ComfyResMultistep` now reads the frozen authority.
- **Budget**: exact additive device term (1,956,118,528 B resident adapter, transformer-load + denoise phases) and a measured host staging term (2× the widest matrix = 33,030,144 B, transformer-load phase, released per matrix). Freeze-time cross-check: budget terms must equal the declaring authority, grid points must be the tier's reviewed count. New fields inherit the per-field hashing/rejection macro guarantees.
- **Envelope**: only the step axis moves, only with an authenticated adapter (9 or 5 vs the pinned 21); every other axis and the no-adapter path are byte-identical, revalidation is never looser than minting (`PublicCompiled { turbo_steps }`).
- **Runtime threading**: `REVIEWED_TURBO_TIERS` table; admission authenticates with no device allocation; transformer-load re-verifies identity/digest/byte counts against what admission froze before materializing deltas. Tier selection is `MOLD_H3_TURBO_ADAPTER` + `MOLD_H3_TURBO_TIER` (both-or-neither), registered per the runtime-env rule; PR 4 replaces the env knob with catalog manifests + policy + CLI/client surfaces.

## Testing

1946/1 (known staged_weights parallel flake, passes isolated) default lib; 2079/1 under h3-private-uat; 17 h3_factory; 235 mold-candle; mold-ai-server h3 failures verified identical on a clean baseline (local weights-on-disk env). fmt + clippy (workspace, h3-private-uat, h3) clean. Negative gates verified in tests: steps 21 with adapter refused, steps 9 without adapter refused.

Real-hardware UAT (8-step render on the RTX 4090 that did the 21-step 1216 s render earlier today, #827 comment) runs next; results will be posted on #827.


jamesbrink:
Codex review fixes landed in `8fe55879`: (1) [P1] the minted Turbo envelope now validates the prepared request — both `None`-defaulting wrapper forms are deleted so every call site must name its step authority (a third broken site at the execution path was found and fixed the same way), with a distinguishing test proving a Turbo envelope accepts its tier and rejects the no-adapter authority; (2) `H3FactoryTurboAdapterAuthority` fields are private and `for_reviewed_tier` is the only constructor — sampler/grid/shift come from `REVIEWED_TURBO_TIERS`, re-proved at validate and at load; (3) load-time transpose transients charged via a new `turbo_adapter_device_staging_bytes` (max module A+B = 20,643,840 B, transformer-load phase only, bound documented); (4) the Metal forced-local dead-code failure fixed per file convention (`cfg_attr` allow) plus a stale import and a latent cfg-gating bug on the `turbo_adapter` binding under h3-private-uat. Full suites green (2084/1 known flake); clippy clean on five feature combos including `--no-default-features`.

Issue #1164, opened 2026-08-19
feat(h3): Metal execution path for compact FL2VA (CorrectnessOnly first)


jamesbrink:
## Current status — audited 2026-09-05

The Metal execution path is already shipped and admitted, and a reduced-size retained H3 Metal render is now qualified; default-resolution execution remains unqualified. The original CUDA-only description was stale: #1183/#1323 added Metal backend/lease validation, portable quantized operations, chunked attention, and the macOS build route. `metal: CorrectnessOnly` remains appropriate.

The ~42.5 GB artifact total is not simultaneous residency. Current Metal Qwen streams one language layer at a time (1,052,855,836 host + 1,191,583,200 device parameter bytes, with staging/workspace charged separately); DiT main blocks also stream. Admission compares the largest phase's combined host/device budget against unified headroom. A 48 GiB fit must be measured rather than inferred from disk sizes. See #1542.

Current-main supports `mold system metal-memory status/set/reset` through the macOS **sysctl** `iogpu.wired_limit_mb`. This is not an ioctl or a process-RSS hard limit. The audited 48 GiB host is at automatic (`0`); no setting was changed. Its installed 0.26.0 binary predates these commands.

## Work remaining

- [x] Metal backend and device-exact lease/placement validation.
- [x] Metal chunked dense attention implementation and deterministic Metal/CPU parity tests.
- [x] Portable INT8 ConvRot and NVFP4-AWQ Qwen path; Metal Qwen language-layer streaming.
- [x] Published correctness-only tier and shipped macOS H3 build route.
- [x] Merge and verify the attention lifetime correction: small real-Metal regression reproduced 135,462,912 retained bytes; preallocated output plus per-chunk completion reduced this to 1,572,864 bytes. Merged in PR #1604 with the INT8 row-lifetime correction. 22 attention tests and the isolated allocation test pass.
- [ ] Complete current phase-budget and guarded real-component qualification on 48 GiB hardware (#1542), preserving a host safety floor and an independent allocation/pressure abort path.
- [x] Retain and inspect a reduced-size conditioned compact FL2VA video with audio on Metal; record allocation samples, host headroom and timing (see evidence below).
- [x] Compare equivalent checkpoint/schedule/conditioning outputs with a CUDA or independently qualified reference; retain numerical and visual evidence. Preserve CUDA's FlashAttention and accelerated quantized paths.
- [x] Update qualification docs and issue completion state from that evidence (PR #1604). Performance and default-resolution qualification remain separate.

[...]

Issue #800, opened 2026-08-07
feat(wan): Metal correctness path for the 1.3B and 5B checkpoints


jamesbrink:
## Motivation
Wan's capability entry declares `metal: Unsupported`, so macOS desktop users (the embedded `metal` engine) and Apple-Silicon `mold serve` hosts cannot run any wan checkpoint — yet wan ships the two best Apple-Silicon fits of any mold video family: `wan21-t2v-1.3b:bf16` (~2.8 GB DiT) and `wan22-ti2v-5b:fp16` (~10 GB), both comfortable in unified memory with the encoder dropped, unlike LTX-2's 19B. LTX-2 just gained a Metal correctness path (#762, closing #597: family-scoped BF16, fused SDPA, framewise VAE decode); wan is architecturally simpler (no block streaming, no audio, Conv2d-slice causal VAE, CPU-seeded noise already guarantees cross-backend seeds). This is genuinely separate work from #597's LTX-2 lineage — new family qualification, referencing that precedent.

## Evidence
- `crates/mold-inference/src/batch.rs:408-431` — `backends: { cuda: Supported, metal: Unsupported, cpu: CorrectnessOnly }`; `seed_contract: CPU_SEED`.
- `crates/mold-inference/src/device.rs:2100-2106` — `gpu_dtype` returns F32 for non-CUDA devices, so a naive Metal run would execute the whole DiT in F32 (2x memory, slow): qualification needs a wan family-scoped BF16/F16 decision like LTX-2's Metal path took.
- `crates/mold-inference/src/wan/model/transformer.rs:582` — attention through the portable `crate::attention` dispatcher; the VAE is the Conv2d-slice decomposition (no conv3d kernel dependency) with feat-cache-bounded decode memory.
- `website/models/index.md:172` — "Wan is CUDA-only for real generation" (to be updated).
- Precedent: commit `313b69e7` (#762) and CLAUDE.md's LTX-2 Metal invariant (checkpoint-backed UAT required before performance qualification).

## Proposed work
1. Qualify `wan21-t2v-1.3b` first: UMT5 encode, DiT forward (fix Metal dtype/op gaps: BF16 RMSNorm/rope/softmax, F64 RoPE table construction on a backend without fp64), FlowUniPC (device-neutral), causal VAE feat-cache decode; choose the wan Metal compute dtype (extend `gpu_dtype` or a family-scoped resolver).
[...]

PR #1183, opened 2026-08-20
feat(h3): CorrectnessOnly Metal execution path

jamesbrink:
Implements #1164's first stage, following the Wan #800 CorrectnessOnly pattern. H3 now compiles and passes its unit suite on real Apple Silicon; the runnable route stays deliberately fail-closed pending 64 GB-class hardware qualification.

## What

1. **Chunked dense attention on Metal** — new `MetalChunkedDenseMath` backend arm: per-chunk QK with full-key softmax (arithmetically identical to unchunked; the unchunked score matrix at released geometry is ~358 TiB). The chunk is a pure function of shape, stored in the frozen plan, hashed into the plan identity (schema 1→2), revalidated at execution. CUDA behaviour unchanged; `correctness_for_device` is the single selector.
2. **`metal_reduce` moved to mold-candle** (mold-inference keeps an alias; four existing call sites untouched) and applied to the audio VAE's `CausalAttention` rank-4 mean — inside candle's Metal mis-reduction window. The visual VAE was audited and is structurally immune (rank-4 Conv2d loops; GroupNorm folds to rank-3).
3. **Explicit INT8 arm selection** — `select_h3_int8_linear_kind` is a pure function (zimage/qwen precedent); Metal takes the portable dequant arm by construction; the native cuBLASLt kernel stays CUDA-only; fp8-scaled weights are refused on Metal by name (NVFP4-AWQ documented exempt — its scales unswizzle to f32 on host).
4. **Metal accepted through the frozen plan** — `H3CandleBackendDevice::Metal`, `compute_capability: Option<(u16,u16)>` where `None` is Metal, lease checks are plan-exact (a CUDA plan handed a Metal lease is refused because placements disagree), `MetalResident` conditioner execution.
5. **Family dtype policy** (`minimax_h3/dtype.rs`, the wan/backend.rs analogue) naming the F32 islands.
6. **Feature graph**: `h3` no longer implies `cuda` — that made an Apple Silicon H3 build inexpressible. `h3-cuda` is the shipping CUDA edge; mold-server's build-script fence gains a separate exact Metal feature set (no NVML, no deviceless "CUDA optional").

## Fail-closed boundary (deliberate)

`h3_admission.rs` still refuses non-CUDA backends, the public profile remains `supported-compact-fl2va-sm89` (a Metal authority carries `None` capability and is refused without needing a Metal branch), and `runtime_available` stays false — advertised as `BackendQualification::CorrectnessOnly`, truthfully a compiled path, not a route. Lifting it needs a 64 GB+ Apple Silicon machine (the compact stack is 42.5 GB unified; the available M4 has 16 GB). A parity-fixture plan against the minimax-h3-mlx oracle is recorded in the series notes.

## Verification

Linux: 243 mold-candle + 94 inference minimax_h3 tests post-rebase, full suites in branch history, clippy/fmt clean, `h3-cuda` check clean, release-contract script PASS. **bender (M4 Mac mini, real Metal): `cargo check --features metal,h3` EXIT 0; 233 mold-candle tests including chunked-plan-vs-CPU equality, never-selects-native-INT8, and fp8 refusal; 85 inference tests including the BF16 dtype policy.** Docs updated (CLAUDE.md/SKILL.md/website model page).


jamesbrink:
Codex review + CI fixes landed (`a367b88f`, `364b39ba`, `d9875ec9`): (1) every SM89 shipping recipe now routes through the `h3-cuda` edge — including two additional instances of the same bug (the desktop crate's own `h3` feature and the qualification-record dev-bin's required-features) — with enforcement moved into the mold-server build script itself (`cuda,h3` now fails the build) plus exact-token recipe guards in the release-contract script, verified to fire; (2) the contradictory Metal + `AssignedCudaThenDrop` pair is refused at construction, and device-resident conditioners must match the frozen backend class; (3) the docs job pins were updated with the truthful Metal wording, and the second pre-existing docs failure (undocumented `MOLD_H3_TURBO_ADAPTER`/`MOLD_H3_TURBO_TIER` from #1178) is fixed by documenting them in the configuration guide. Full gate set green including all four contract scripts and a device-free `h3` check.

PR #1184, opened 2026-08-20
feat(tui): Library titles and trash, plus overwrite guards on every trash move

jamesbrink:
## Summary

TUI surface of **Library organization** (stacked on #1173/#1175) plus a cross-surface data-integrity fix from its codex review:

- **TUI Library**: lists only live rows (`MetadataDb::list_live` — trashed prints disappear from the grid), shows print titles in the Details panel and full detail view, and `d` moves to the trash — locally via `mold_db::trash` (tombstone + row flag), remotely via `MoldClient::trash_gallery_image` — with capability-gated confirm copy that degrades to honest hard-delete wording when any owning machine cannot trash (`RemovalKind`). Settings gains a `Trash (days)` row editing `gallery.trash_retention_days` through the shared config writer. Key→action contract unchanged.
- **Overwrite guards (codex P1)**: `fs::rename` replaces an existing destination on Unix, so trashing a live same-name replacement could silently destroy previously trashed bytes. Every trash move now refuses same-name collisions instead: the server's `move_gallery_file_to_trash` / `move_gallery_file_from_trash` / `restore_trashed_archive_filename` and the TUI's local move — each pinned by a regression test proving the retained bytes survive.
- TUI help overlay wording made honest for the permanent-delete fallback (codex P2).
- Docs: `website/guide/tui.md`, CLAUDE.md TUI sentence, changelog fragment.

## Test plan
- [x] `cargo test -p mold-ai-tui` (689), `cargo test -p mold-ai-server --lib` (1568 pass; the 2 known 1 s-timeout load-flakes pass in isolation)
- [x] `cargo fmt --all -- --check`, `cargo clippy -p mold-ai-tui -p mold-ai-server --all-targets -- -D warnings`
- [x] Codex peer review (P1 + P2 addressed in 08a7315c)
- [ ] CI


jamesbrink:
The Metal forced-local typecheck failure here is not this PR's — main currently fails that check (inherited from #1178, which merged at its pre-review head). The fix is in #1180 (cfg_attr dead-code allows on the always-compiled Turbo accessors); once it merges, a rebase/rerun here goes green.

PR #1196, opened 2026-08-20
fix(h3,video): Turbo terminal provenance + silent H.264 frame skips

jamesbrink:
## Summary

Two terminal-gate defects found by running the 8-step Turbo render on the RTX 4090 — each cost a complete render (804 s and 931 s) rejected at the last validation.

### 1. Sampler provenance hard-coded RES-multistep

A reviewed Turbo attempt runs `ComfyEuler`, but `validate_private_sampler_provenance` compared against a hard-coded `ComfyResMultistep` → `private H3 sampler provenance differs from the prepared attempt authority` after a complete render.

- `H3PrivateFl2VaOwnerFacts` gains `sampler: H3SamplerKind`, set from `frozen_factory.quantization().sampler_kind()` — the same authority the runtime executes (`pipeline.rs:775-781`, provenance at `:915`).
- Owner validation requires a comfy-simple-grid integrator; the terminal gate compares against the owner's field.

### 2. Silent H.264 frame skips shorten the muxed track

With fix 1 in place the rerun failed at the next axis: `terminal media provenance mismatch: echo-duration`. Root cause: `Mp4StreamEncoder` left openh264's `bEnableFrameSkip` at its default (on) — under rate-control pressure a pushed frame emits an empty bitstream and `push()` silently dropped it from the sample table, so mdhd duration lands below the rendered frame count. Content-dependent: the 21-step render passed, the 8-step Turbo output tripped it. LTX and Wan clips would have shipped the shortened track with no error at all.

- `skip_frames(false)` (matches the LTX-2 preprocess round-trip encoder's existing setting).
- A pushed frame yielding no picture NALs now fails the encode loudly.
- New test pins video-track sample count, timescale, and mdhd duration to exactly frames/fps.

Also fixes a test-only compile error under `h3-cuda,mp4`: the `private_qwen` CUDA smoke lease still built `H3CandleBackendDevice::Cuda` with an `Option`-wrapped compute capability after #1189 made the variant exact.

## Tests

- Extended `terminal_sampler_provenance_is_exactly_bound_to_the_owner` (comfy-euler refused for a non-Turbo owner), new `turbo_owner_accepts_only_the_frozen_euler_sampler`, new `owner_facts_refuse_a_sampler_off_the_comfy_simple_grid`, new `mp4_track_duration_is_exactly_frames_over_fps`.
- `cargo clippy --workspace --all-targets`, `clippy -p mold-ai-inference --features h3-cuda,mp4` and `--features mp4`, `cargo fmt --check` all clean; video_enc (20) and preprocess (16) suites pass.
- Codex review of fix 1: no findings ('sampler authority is correct… no missed constructors').

The 6 pre-existing stale Embedded-path `private_server` tests under `--features h3` remain failing before and after (unrelated, tracked).


jamesbrink:
Live UAT on the RTX 4090 confirms both fixes: with this branch's binary the 8-step Turbo render **published** — 759.5 s generation, 124-frame MP4 + synchronized AAC, validated with ffprobe (vs 1216.1 s for the 21-step baseline). The two prior attempts failed exactly at the axes this PR fixes (sampler provenance at 804 s, echo-duration at 931 s). Full numbers on #827.

PR #1199, opened 2026-08-20
feat(h3): first-class Turbo tier manifests, replacing the env-var knob

jamesbrink:
## Summary

The reviewed MiniMax H3 Turbo LoRA tiers become ordinary model identities: **`minimax-h3-fl2va:comfy-pruned-int8-turbo-8step`** (9 terminal-inclusive steps) and **`minimax-h3-fl2va:comfy-pruned-int8-turbo-4step-768p`** (5 steps). Ref2VA's reviewed tier deliberately gets no tag while Ref2VA execution is unimplemented.

- **One authority table** `mold_core::minimax_h3::REVIEWED_TURBO_MANIFEST_TIERS` pins tier id, adapter file, size, sha256, and reviewed steps; a contract test pins it byte-for-byte to mold-candle's `H3TurboLoraTier`. Each turbo manifest is the exact base FL2VA file set plus one `ModelComponent::DistilledLora` stored at `shared/minimax-h3/loras/` (shared by both tags); `loras/*` pin their own HF revision (`dc559027`) while the base stack keeps the reviewed revision.
- **Env knob demoted**: `MOLD_H3_TURBO_ADAPTER`+`MOLD_H3_TURBO_TIER` override only under `h3-private-uat`; in every other build a set env pair is a hard error (contradictory authority with a tag; 'UAT-only, use the tag' without one) — never silently ignored. `resolve_turbo_selection` is the single rule for admission and transformer load; `for_reviewed_tier` remains the sole authority constructor.
- **Engine partition stays `FL2VA_COMFY`** (evidence/storage/qualification untouched); request/media/provenance carry the full turbo id, paired to the frozen adapter by `media_model_matches_h3_authority` — this also fixes a latent terminal-gate mismatch in `prepare_h3_private_fl2va_attempt`.
- **Capability wire**: additive `turbo` list on the fl2va partition (older clients hard-fail on duplicate task partitions); `/api/models` gains one row per executable variant with its own steps and disk.
- **Policy**: both reviewed-ID allowlists derive from `REVIEWED_COMPACT_MODELS`; unreviewed turbo lookalikes stay fail-closed.
- **CLI/Discord defaults bug**: H3 no longer submits the official-BF16 `DEFAULT_STEPS` (50) — steps default from the model's manifest (21/9/5), dimensions pinned at 1344×768. `studio` derives expected steps from `MINIMAX_H3_REVIEWED_FL2VA_STEPS` instead of hard-coding 21. TUI needed no code.
- Docs: model guide variants table + Turbo section, configuration env rows demoted to UAT-only, README/CLAUDE.md/SKILL.md, verify-docs pins, changelog fragment.

## Context

The 8-step tier published on an RTX 4090 at **759.5 s vs 1216.1 s** for the 21-step baseline (#827) — this PR makes that runnable as `mold run minimax-h3-fl2va:comfy-pruned-int8-turbo-8step` with a plain `mold pull`.

## Gates

fmt/clippy (workspace + `h3-cuda,mp4`)/`check --workspace --all-targets`/default-features check all clean. mold-ai-core 1256 ✓, inference 1965 ✓ (+known staged_weights flake), server hermetic h3 103 ✓ + non-route 142 ✓, CLI 589 ✓ (+11 known local-env), discord 143 ✓, TUI 688 ✓, studio 934 / web 1475 / desktop 4259 ✓. Known pre-existing failures unchanged from main.


jamesbrink:
Codex adversarial review at 8ffe9eec: no P1/P3; the sole P2 (Prettier formatting on three files) is fixed in 0d4cc023. Verified paths: both allowlists derive from exact REVIEWED_COMPACT_MODELS membership with cv:/hf: checked before and after metadata resolution; a turbo tag resolves exactly one pinned adapter/tier with the digest authenticated at admission AND independently at transformer load (cross-tier substitution fails); ordinary h3 builds reject the env pair before any weights load; the capability addition is optional in Rust and the shared Studio parser; media_model_matches_h3_authority + the GPU-worker publication contract bind prepared facts, request, response, and terminal echo.

PR #1200, opened 2026-08-20
fix(h3): turbo tags share the base model's storage instead of re-downloading 41 GB

jamesbrink:
## Summary

Live UAT of #1199 on the RTX 4090: `mold pull minimax-h3-fl2va:comfy-pruned-int8-turbo-8step` on a machine that already had the base model began re-downloading the entire 20.97 GB transformer (and the rest of the ~41 GB base stack) into a new `minimax-h3-fl2va-comfy-pruned-int8-turbo-8step/` directory. Only the 1.96 GB adapter is genuinely new.

**Root cause:** `mold_core::manifest::storage_path` keys model-specific components on the manifest's own name, so each turbo tag claimed its own directory and never saw the base install.

**Fix:** H3 manifests resolve their model-specific directory through `mold_core::minimax_h3::storage_identity(name)` — reviewed turbo tags map to `FL2VA_COMFY`, everything else to itself. The turbo tags' base-stack files land on the exact paths the base install owns (the adapter already lived in `shared/minimax-h3/loras/`). Existing installs need no migration. Everything downstream derives from `storage_path`, so one change point covers download-verify, repair, load, artifact contracts, and inventory.

**Semantics:** pull turbo with base installed → base files sha-verify in place, only the adapter downloads; pull base after turbo → nothing downloads. Removal was already path-refcounted (`removal.rs`): removing a turbo tag deletes only its adapter while the base remains; removing the base with a turbo tag installed keeps shared files and names the tag in `used_by`. Per-model disk sizes follow the same overlap convention as FLUX's shared T5.

**Tests (failing first):** `turbo_manifests_store_the_base_stack_in_the_base_models_directory` pins every non-adapter turbo file to the base manifest's exact storage path; `turbo_removal_refcounts_protect_the_shared_base_stack` pins both removal directions.

## Gates
fmt, workspace clippy, h3-cuda clippy, generation-profiles --check, website prettier + docs verifier all clean. With MOLD_* env unset: mold-core 1258, CLI 600/600, server h3 103, inference minimax 102, TUI 688, discord 143. The 3 `--features h3` mold-core failures reproduce byte-identically on pristine main (pre-existing).

### DIFF
--- a/changelog.d/h3-turbo-shared-storage.md
+++ b/changelog.d/h3-turbo-shared-storage.md
@@ -1,3 +1,5 @@
-### Fixed
-
-- MiniMax H3 Turbo tags now store their base stack in the base checkpoint's directory: pulling a Turbo tag on a machine with `minimax-h3-fl2va:comfy-pruned-int8` installed downloads only the ~1.96 GB adapter instead of re-downloading ~41 GB into a tag-named copy, and removal ref-counting protects the shared files in both directions.
+- **MiniMax H3 Turbo tags share the base checkpoint's storage.** Pulling a
+  Turbo tag on a machine with `minimax-h3-fl2va:comfy-pruned-int8` installed
+  downloads only the ~1.96 GB adapter instead of re-downloading ~41 GB into a
+  tag-named copy, and removal ref-counting protects the shared files in both
+  directions.
--- a/changelog.d/h3-turbo-shared-storage.md
+++ b/changelog.d/h3-turbo-shared-storage.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- MiniMax H3 Turbo tags now store their base stack in the base checkpoint's directory: pulling a Turbo tag on a machine with `minimax-h3-fl2va:comfy-pruned-int8` installed downloads only the ~1.96 GB adapter instead of re-downloading ~41 GB into a tag-named copy, and removal ref-counting protects the shared files in both directions.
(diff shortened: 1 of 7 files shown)


jamesbrink:
Codex round 2 resolution (f52289ce):

**P2 (half-installed ownership) — fixed.** `build_ref_counts` skips any configured entry missing any of its own files on disk, deliberately NOT the suggested `manifest_model_is_downloaded` filter — that resolves manifest storage paths and would strip ownership from configured non-manifest models (custom/`cv:` installs with explicit paths), deleting shared files complete installs still use. Path-existence of the entry's own files is the completeness test that works for every entry shape. The TUI's silently-diverged private `build_ref_counts` copy now delegates to the core rule.

**P1 (phantom base) — presence-based inventory and ownership KEPT, as a deliberate documented decision.** A Turbo-only install and an explicit base+Turbo install are byte-identical on disk (pulls write no config entry; no receipt exists), so any rule deleting the stack on Turbo removal in the first scenario also deletes an explicitly pulled base in the second. Between a visible, removable, truthfully-reported 41 GB and the risk of destroying 41 GB someone wanted, keeping wins: the base IS complete/sha-verified/runnable, removal reports `kept: used by minimax-h3-fl2va:comfy-pruned-int8`, full cleanup is two removals, and `AlreadyAvailable` on a base pull stays correct. Documented on `storage_identity`'s doc comment and in the model guide.

Three new tests: `incomplete_configured_models_do_not_own_shared_files`, the half-installed-Turbo removal case, and the end-to-end turbo-only scenario (`turbo_only_install_keeps_the_base_stack_until_the_base_is_removed`). Confirmed-good from round 1 unchanged: sha-verified reuse, exact-tag download dedup, H3-only scoping.

Issue #1204, opened 2026-08-20
h3-private-uat: 10 deterministic gpu_worker claimed_h3_*/claimed_ref2va_* test failures (pre-#1201)


jamesbrink:
Running `cargo test -p mold-ai-server --features h3-private-uat --lib` correctly (namespaced per the NFS-automount note, hermetic env, single-threaded name-diffed to exclude load flakes) yields **10 deterministic failures**: `gpu_worker::tests::claimed_h3_*` (7) and `claimed_ref2va_*` (3). Representative: `fake Ref2VA success: "MiniMax H3 prepared attempt changed its owner run binding"`.

Provenance, measured: the failing set is byte-identical on `dc52dd51` — the commit BEFORE #1201 — so it predates the Ref2VA admission threading, the frozen-plan slice, and every Turbo PR. Not concurrency (same set at `--test-threads=1`) and not the known local-config leakage (same set with HOME/XDG isolated to empty dirs).

Two questions to settle:
1. Is the `h3-private-uat` feature combination covered by any required CI gate? (If it were, these would not have survived.)
2. Is the owner-run-binding fence genuinely broken for the fake-success test path, or do the tests encode a stale contract from before the prepared-attempt identity work?

Load-dependent flakes seen alongside (routes_test gallery publication, placement_preview ×2, partial_transport_failure, all_temporarily_low_devices) all pass in isolation and are NOT part of this set.

jamesbrink:
## Confirmed still present on `6d46e55f` — 2026-08-21

Re-ran with `HOME`/`XDG_CONFIG_HOME` isolated to empty scratch directories, to rule out this machine's known local-`config.toml` leakage:

```
$ cargo test -p mold-ai-server --features h3-private-uat --lib -- --test-threads=1 claimed_
test result: FAILED. 10 passed; 10 failed; 0 ignored; 1623 filtered out; finished in 0.29s
```

The failing set is byte-identical to the one this issue lists — 7 `claimed_h3_*` plus 3 `claimed_ref2va_*`. Representative failure is still the one quoted here:

```
thread 'gpu_worker::tests::claimed_h3_fake_success_commits_and_publishes_once_without_touching_cache'
panicked at crates/mold-server/src/gpu_worker.rs:6390:47:
fake H3 success result: "MiniMax H3 prepared attempt changed its owner run binding"
```

[...]

Issue #825, opened 2026-08-07
feat(h3): implement ordered multimodal Ref2VA


jamesbrink:
Part of the MiniMax H3 native audio-video milestone (tracking: #814). Add the second released task partition: semantically ordered image/video/audio reference-to-video-and-audio generation.

## Conditioning implementation

- [x] Consume the authoritative ordered `GenerationReference` list from #837 without regrouping by media type. ComfyUI's current UI grouping is not the model contract.
- [x] Build exact Qwen `<Picture i>`, `<Video i>`, and `<Audio i>` labels and token/media presentation in request order.
- [x] Encode images, video frames, video soundtracks, and standalone audio with their documented per-modality preprocessing.
- [x] Place a video's soundtrack immediately before its visual rows and give the associated blocks the correct shared rotary origin.
- [x] Pack `[text | reference blocks in request order | generated audio | generated video]`, modality IDs, per-row time values, and MM-RoPE positions exactly.
- [x] Preserve the Ref2VA transformer partition identity across download, queue, cancellation, retry, gallery metadata, templates, and “use as prompt/source” flows.
- [ ] Ref2VA **execution** on a shipping build — `private_server.rs` still answers `Task::Ref2va => false` outside `h3-private-uat`, and no release recipe carries that feature.
- [x] Truncate reference audio to the generated duration only where the official contract requires it; preserve user-visible validation of original/used durations.

## Full released input matrix

- [x] Image-only references, including multiple ordered subjects/styles.
- [x] Video-only references with and without soundtrack.
- [x] Image plus standalone audio.
- [x] Video plus additional standalone audio.
- [x] Mixed image/video/audio orderings up to the 9-image/3-video/3-audio/12-file limits.
- [x] Rejection of audio-only input, over-duration aggregates, unsupported codecs, corrupt streams, and preprocessed sequences that exceed admitted packed-row limits.

## Product semantics

- [x] Label the mode as reference-guided semantic resynthesis; do not imply source-frame preservation or expose a denoise-strength control.
- [x] Show one-based reference order, video soundtrack association, per-file duration, and limit budget before submission.
- [x] Store reference kind/order/name/hash/duration provenance without retaining generation base64 or leaking server-local paths.

## Acceptance

- #832 and CUDA UAT cover image-only, video-with-soundtrack, image+audio, two order-swapped mixed lists, and every limit failure.
- Swapping two references changes labels/coordinates/fingerprint and produces correspondingly different saved order metadata.
[...]

PR #1205, opened 2026-08-20
feat(h3): Ref2VA-shaped frozen plans from the admitted task

jamesbrink:
## Summary

Final model-side slice for #825: carries the Ref2VA prepared request through `build_private_fl2va_admission_attempt` to a Ref2VA-shaped frozen plan. Four remaining FL2VA-pinned points now follow the admitted task: the storage authority's manifest resolution, the transformer contract validator, the VAE contract validator, and the raw checkpoint's expected artifact. The reopen path follows what admission froze, so a re-derived attempt cannot drift to the other task's checkpoint.

**One authority, pinned:** reconciling with #1203 revealed the slice was carrying a second copy of the task→partition mapping — the same duplication class that produced the turbo qualification defect — so both call sites now delegate to `contract::base_compact_model_for_task`, and a test walks all four identities (both bases + both turbo tags) asserting the storage authority's task-derived identity equals the route's partition and is the manifest actually opened.

**Storage fact as a test:** the Ref2VA stack differs from FL2VA in exactly two manifest entries — the transformer weight file and the `transformer_ref/config.json` lookup key (content and sha256 identical, 546 bytes, 74c11bff…); Qwen weights and both VAEs shared, asserted by digest comparison.

**FL2VA byte-identity, verified:** a prepared-attempt identity pin (862ae3c2…) captured BEFORE this work still holds across #1200/#1201/#1202/#1203 and this slice — now a permanent regression test. The cross-task validator test checks both directions.

Ref2VA remains campaign-build-only under `h3-private-uat`; the public allowlist is untouched (`reviewed_allowlist_is_scoped_to_fl2va` passes on both feature sets). Next: the instrumented capture campaign on the 4090.

## Known pre-existing failures (NOT this PR)

Running the `h3-private-uat` server suite correctly surfaces 10 deterministic `gpu_worker::tests::claimed_h3_*`/`claimed_ref2va_*` failures that are byte-identical on the commit before #1201 — tracked in #1204; this branch adds zero to that set.

## Gates
fmt, `check --workspace --all-targets`, `h3-cuda,mp4` clippy, Metal-proxy clippy all clean; `private_opened_evidence` 13, `h3_factory` 34, inference `h3-private-uat,mp4` 2166 ✓ (+known staged_weights flake).

### DIFF
--- a/crates/mold-inference/src/h3_factory.rs
+++ b/crates/mold-inference/src/h3_factory.rs
@@ -5600,21 +5600,51 @@ mod tests {
 
     /// FL2VA's frozen plan must be byte-identical across the Ref2VA work.
     ///
-    /// The prepared-attempt identity covers the request, the raw checkpoint,
-    /// and the whole target budget, so pinning it to a literal digest is the
-    /// cheapest complete statement that adding a second task changed nothing
-    /// about the first. Recomputed from the fixture, which is path- and
-    /// machine-independent by construction.
+    /// Two pinned literals, each with its own coverage:
     ///
-    /// If this digest changes, an FL2VA frozen plan changed. That is either a
-    /// deliberate FL2VA change — update the literal and say so — or a Ref2VA
-    /// change that leaked, which is a bug.
+    /// * The prepared-attempt digest (862ae3c2…) is
+    ///   `expected_h3_factory_prepared_attempt_identity` over the
+    ///   hand-assembled fixture, so it covers the identity FUNCTIONS — the
+    ///   request, endpoint, raw-checkpoint, and target-budget hash layouts and
+    ///   every field they bind — but NOT the production budget builder
+    ///   (`build_canonical_private_fl2va_target_budget`), which needs opened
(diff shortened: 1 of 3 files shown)


jamesbrink:
Codex findings addressed in c8973fc2: (1) storage revalidation, VAE planning, and task-config opening all key on `comfy_task_sources(self.task)` / `base_compact_model_for_task` — a Ref2VA authority now survives its own revalidation with a distinct identity; (2) the reopen pre-storage gates are one task-keyed seam (`validate_frozen_reopen_route`) driven by the FROZEN task, with tests proving a frozen Ref2VA attempt reaches and passes the task-aware validators, a crossed FL2VA request is refused, and Ref2VA stays refused outside the campaign build; (3) the identity pin now also routes through the production `FrozenH3FactoryAuthority::new_contract_only` builder and pins the backend-plan identity (b9e5fe2e…) additively — the original 862ae3c2… digest did not move, and the comment states exactly what each pin covers. Inference `h3-private-uat,mp4` suite: 2171 passed / 0 failed.

PR #1208, opened 2026-08-20
fix(h3): derive the capture-scope provisional ceilings from the corrected ledger

jamesbrink:
## Summary

Campaign attempt 5 on the RTX 4090 was refused at the admission floors: the capture-scope profile demanded **51.27 GB device** before any request-specific term — over double the card's VRAM — because `capture_runtime_bounds()` carried the PRE-correction FL2VA sizing (attention 10.13 GB×5 'quadratic', FFN 15.30 GB×3) that the measurement-grounded ledger corrections had already superseded (observed 6.17 / 7.64 GB, #827). The ×5 quadratic term sized a score matrix the FlashAttention v2 route never materializes.

Every ceiling is now derived term-by-term from the corrected observations, scaled to the campaign envelope by the dimension that actually drives each term (sequence-linear for attention/FFN at row ratio 88,334/39,776; canvas-pixel ratio for the condition VAE; 2× the corrected public ceiling for Qwen activation, sized to small ordered sets — the twelve-still envelope maximum would blow host headroom beside the 19.07 GB CPU-placed conditioner and is refused by the exact per-attempt check instead of pre-granted). Provenance table in the PR thread; formulas in the `capture_runtime_bounds` doc comment with observed constants isolated in an `fl2va_observed` module cross-referencing `public_runtime_bounds`/#827.

**Resulting floors: device 17.58 GB (clears the 4090's 24.97 GB sample by 7.4 GB), host 19.87 GB.** The observed/ceiling capture report is untouched — the reviewer still transcribes observations, never ceilings.

Tests: the provisional-profile test now requires the precheck to admit attempt 5's exact device/host sample and pins both floors and the sequence-scaled ceilings to their formulas; the floor-invariant sweep covers the derived capture record.

## Gates
fmt, `check --workspace --all-targets`, clippy (inference `h3-cuda,mp4` + `h3-private-uat,mp4`, mold-ai set, mold-ai-server `h3-private-uat`), release-contract script PASS, inference uat suite 2175 ✓ (+known flake), Metal clippy run clean on Apple Silicon at this SHA.

### DIFF
--- a/crates/mold-inference/src/minimax_h3/private_opened_evidence.rs
+++ b/crates/mold-inference/src/minimax_h3/private_opened_evidence.rs
@@ -1570,6 +1570,64 @@ fn validate_vae_contract(expected_task: Task, task: Task, canonical_model: &str)
     Ok(())
 }
 
+/// FL2VA's corrected observed Qwen activation workspace and the Qwen
+/// sequence it was measured over (2026-08-19 on hal9000, RTX 4090 SM89,
+/// issue #827): the qualifying render encoded 1,058 output-text rows plus
+/// 4,032 vision rows. `public_runtime_bounds` applies its margin policy to
+/// the same observation; restated here because that function is compiled
+/// only under `h3`.
+const FL2VA_OBSERVED_QWEN_ACTIVATION_WORKSPACE_BYTES: u64 = 3_400_171_520;
+const FL2VA_OBSERVED_QWEN_SEQUENCE_ROWS: u64 = 1_058 + 4_032;
+
+/// The Qwen activation workspace one request charges into its exact budget
+/// and freezes into its factory authority.
+///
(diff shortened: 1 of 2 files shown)


jamesbrink:
Codex P2 fixed in 28bb0692: the Qwen activation charge is now request-derived through one seam (`qwen_activation_workspace_demand_bytes`) used by both charging sites — FL2VA passes the reviewed grant verbatim (byte-identical), Ref2VA derives demand from the request's actual packed rows under the same ×1.15 + 64 MiB policy as the corrected public ceilings, and a demand above the profile's provisional grant is a named refusal carrying both figures. The 2× constant survives only as the precheck/profile grant ceiling. The 3-reference max-canvas shape (13,346 rows, 2.62×) derives 10.27 GB and is refused by name; the scripted campaign shape stays admitted at exactly the corrected public ceiling. Full uat suite 2177/0.

PR #1209, opened 2026-08-20
fix(ltx2): noise soft appended conditioning

### LINKED ISSUE #1080: audit(ltx2): chain soft-anchor appended tokens initialize un-noised under a soft timestep
Follow-up from #1055 (found during the PR #1076 audit, deliberately out of scope there).

Mold's smooth-chain continuation appends the previous clip's anchor frame as a soft appended condition (`CHAIN_SOFT_ANCHOR_STRENGTH = 0.4`, `ltx2/pipeline.rs`). Appended tokens initialize to their **clean values** (`apply_appended_video_conditioning` pushes the full tokens; the denoise mask expresses the 0.4 strength), so the first transformer pass sees a token containing **zero noise** under a per-token timestep claiming `0.6·σ` — the same latent/timestep-mismatch class #1055 fixed for replacements, in the opposite direction ("too clean" rather than "too sourced").

Upstream's noiser (`noisers.py:32-33` @ fd4ded7) lerps appended keyframe tokens against fresh noise exactly like replacement tokens, so upstream a 0.4-strength appended token would start as `0.4·C + 0.6·N`.

This was NOT changed in #1076 because:
- every production appended condition except the chain anchor is strength 1.0 (keyframes, IC-LoRA/lip-dub references), where mold and upstream agree exactly;
- the chain soft anchor is a mold-specific mechanism (not an upstream concept) whose 0.4 was tuned against the current initialization, and changing it risks regressing chain seam quality;
- fixing it needs a dedicated salted noise stream for appended token shapes (`request_seed ^ SALT` via `seeded_randn`, CPU-generated for cross-backend determinism) plus `reapply_stage_video_conditioning` carrying the evolved slice for soft appended tokens instead of re-pinning source every step.

Work items:
1. Add the appended-noise path behind the existing init seam (`initialize_video_stage_latents`).
2. Re-tune/validate `CHAIN_SOFT_ANCHOR_STRENGTH` against the corrected init with seeded 2-clip seam UAT.
3. Keep `reapply` re-pinning for hard (s ≥ 1.0) appended tokens; never drop tokens (count-desync invariant).

Witness tests to start from: `appended_tokens_initialize_to_their_clean_values` (documents today's behavior) and `soft_appended…` sketch in the #1055 plan.

jamesbrink:
## Summary

- align soft appended LTX-2 conditioning with the upstream single-blend noiser semantics
- keep soft appended tokens evolving during denoising while hard keyframes remain pinned
- add deterministic seeded-noise and reapplication regression coverage
- retain the 0.4 chain anchor after seeded two-clip seam UAT
- unblock Rust 1.98 CI by applying semantics-preserving `drain_collect` and constant-chunk lint migrations present on current main

## Verification

- `nix develop -c cargo test -p mold-ai-inference ltx2::runtime::tests` (142 passed)
- focused soft initialization, soft reapply, and hard keyframe tests passed
- `nix develop -c cargo clippy --workspace --all-targets -- -D warnings`
- `nix develop -c cargo test -p mold-ai-catalog` (189 passed, 2 ignored)
- LTX-2 preprocess tests (13 passed, 1 measurement ignored)
- MiniMax H3 processor tests (6 passed)
- private H3 audio/Qwen capture bins checked with `dev-bins,h3-private-uat`
- `git diff --check`
- CUDA LTX-2 19B distilled seeded two-clip UAT: both seams visually smooth; frame 48→49 mean absolute luma delta 3.29458 old vs 3.26927 fixed

The full inference crate run passed 1966 tests with 8 ignored and one environment-only Metal discovery failure because that package test build did not enable Metal; all changed LTX-2 tests passed.

Upstream noiser remains a single clean/noise interpolation: https://github.com/Lightricks/LTX-2/blob/main/packages/ltx-core/src/ltx_core/components/noisers.py#L30-L36

Closes #1080

### DIFF
--- a/changelog.d/ltx2-soft-appended-noise.md
+++ b/changelog.d/ltx2-soft-appended-noise.md
@@ -1,0 +1,1 @@
+- **LTX-2 chain anchors now start at the noise level their soft timestep declares.** Smooth multi-clip continuations preserve the intended evolving identity anchor instead of repeatedly pinning its appended token clean ([#1080](https://github.com/utensils/mold/issues/1080)).
--- a/crates/mold-candle/src/minimax_h3/comfy_dit.rs
+++ b/crates/mold-candle/src/minimax_h3/comfy_dit.rs
@@ -3104,7 +3104,7 @@ mod tests {
         let tensor = &header[name];
         let start = tensor["data_offsets"][0].as_u64().unwrap() as usize;
         let end = tensor["data_offsets"][1].as_u64().unwrap() as usize;
-        for bytes in data[start..end].chunks_exact_mut(4) {
+        for bytes in data[start..end].as_chunks_mut::<4>().0 {
             bytes.copy_from_slice(&value.to_le_bytes());
         }
     }
--- a/crates/mold-catalog/src/live.rs
+++ b/crates/mold-catalog/src/live.rs
@@ -878,7 +878,7 @@ async fn civitai_search_paged(
             // the walk resumes from the stored cursor on the next request
             // and `total()` keeps advertising more rows either way.
             if chain.next_page == opts.page && !chain.leftover.is_empty() {
-                let entries: Vec<CatalogEntry> = chain.leftover.drain(..).collect();
+                let entries = std::mem::take(&mut chain.leftover);
                 chain.emitted += entries.len();
                 let page_result = LiveSearchResult {
                     entries,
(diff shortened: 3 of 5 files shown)


jamesbrink:
Your Metal forced-local failure on fdc0def5 is not your branch: the Mac runners' stable toolchain moved to Rust 1.98 and its new `chunks_exact_to_as_chunks` lint fires in `mold-candle/src/minimax_h3/processor.rs` (H3 code). Fixed on main in 806d9b6c — rerun the failed job (or push any commit) and the merge-ref build will pick it up.

jamesbrink:
Second 1.98 lint site fixed on main too (24decc66 — swept all five remaining constant-size `chunks_exact` calls workspace-wide, including the `chain_job_runner.rs:3736` one your `rust` job hit). Rerun failed jobs after the current attempt settles and the merge ref will be clean.

jamesbrink:
Third 1.98 lint site fixed on main (iter_mut().next() on arrays in execution_equivalence.rs — your latest rust failure). Rerun and the merge ref should finally be clean; these were all toolchain-bump fallout, none of them your branch.

Issue #1221, opened 2026-08-21
feat(flux): PuLID cross-attention adapter across every FLUX transformer variant


jamesbrink:
Part of milestone **PuLID-FLUX: functional**. Depends on #1219 and #1220 (which may land in parallel). This single mold PR bumps the candle revision + `Cargo.lock`, wires `forward_with_hook`, and implements the adapter. All four transformer variants are wired and unit-tested, but milestone 1 QUALIFIES and advertises only `flux-dev:q4` and `flux-dev:q8` (the tier bare `flux-dev` resolves first); bf16, Krea, offload-under-pressure, LoRA, and img2img combinations are qualified in milestone 2.

## Reference
- `ToTheBeginning/PuLID` `flux/model.py` (`pulid_double_interval = 2`, `pulid_single_interval = 4`; `img = img + id_weight * self.pulid_ca[ca_idx](id, img)` in the double loop; `real_img = real_img + id_weight * self.pulid_ca[ca_idx](id, real_img)` on the image slice in the single loop) and `pulid/encoders_transformer.py::PerceiverAttentionCA`.
- Executable oracle: `tmp/sdcpp` (`src/model/adapter/pulid.hpp`, `src/model/diffusion/flux.hpp:993-1004` count, `:1120-1161` injection; `docs/pulid.md`). `build/bin/sd-cli` is already built.

## Scope
- `crates/mold-inference/src/flux/pulid.rs`: `PerceiverAttentionCA` — `norm1` (id tokens), `norm2` (image tokens), `to_q` 3072→2048, `to_kv` 2048→4096 split, 16 heads × 128, `to_out` 2048→3072; attention via `crate::attention` (Metal auto-chunk rules apply). `PulidAdapter { ca: Vec<_> }` loaded from `pulid_ca.{i}.*` (fp16 → working dtype), count `ceil(depth/2) + ceil(depth_single/4)` = 20 for FLUX.1, asserted against tensors present. Modules are always resident — never streamed with the offloaded blocks.
[...]

Issue #1220, opened 2026-08-21
feat(core): PuLID request contract, identity capability, and asset dependencies


jamesbrink:
Part of milestone **PuLID-FLUX: functional**. May proceed in parallel with #1219; #1221 depends on both.

### Wire contract (`crates/mold-core/src/types.rs`)
Additive `GenerateRequest` fields, modelled on `source_image` / `source_image_name` (`types.rs:1173-1188`), NOT on H3's `references` (Ref2VA-specific, `types.rs:646-705`):
- `id_image: Option<Vec<u8>>` (base64, PNG/JPEG), `id_image_name: Option<String>` (provenance only)
- `id_weight: Option<f64>` (default 1.0, validated `0.0..=3.0` — upstream UI range; sd.cpp documents 0.7–1.2 as typical)
- `id_start_step: Option<u32>` (default 0, must be `< steps`)

`validate_request`: **milestone 1 accepts identity only for `flux-dev:q4` and `flux-dev:q8`** (the tier bare `flux-dev` resolves first). Every other model/tier — bf16, Krea, flux2, schnell, other families — is rejected with a message naming the supported models; the gate widens in milestone 2 as tiers are qualified. Compatible with neither LoRA nor img2img in milestone 1 (rejected explicitly, qualified in milestone 2).

**Bounded decode:** define named limits for `id_image` — max encoded bytes, max per-axis pixels, max decoded pixels, max decode allocation — and reject unsupported magic bytes and every limit BEFORE any unbounded allocation. (`validation.rs:1881-1885` is only the existing PNG/JPEG sniff precedent; it does not supply these limits.) Boundary and limit-plus-one tests.

`OutputMetadata`: `id_image_sha256`, `id_image_name`, effective `id_weight`, effective `id_start_step` (mirror `source_image_sha256`, `types.rs:2052`). CLI/TUI metadata parsers tolerate absence.

### Capability
Identity support lives in `GenerationCapabilitiesProfile` (`crates/mold-core/src/generation_profile.rs:263`); the additive per-model `/api/models[].supports_identity` derives from that same authority and is true only when the `pulid` feature is compiled AND the model is a qualified identity. Compile-valid `pulid` feature edges (`mold-inference` → `mold-server` → `mold-cli`) are established in this PR; #1223 decides default/release enablement.

### Assets (`crates/mold-core/src/manifest.rs`)
[...]

PR #1241, opened 2026-08-21
fix(core): gate identity on runtime readiness, pin dependency hashes, and freeze the license text

jamesbrink:
Follow-up to #1239 (merged before its codex review fixes landed). Addresses all three codex findings on #1220:

- **P1 — never advertise a capability no engine executes.** `supports_identity` (profile + `/api/models[]`) and identity admission now require `identity::identity_runtime_available()` = the `pulid` feature AND a single `IDENTITY_RUNTIME_READY` constant (false until the adapter PR #1221 flips it in the same change that makes the FLUX engine consume `FrozenEngineConfig.identity_assets`). The feature-off and adapter-pending refusals are distinct messages so clients send operators to the right fix. Tests are correct for both values of the constant.
- **P1 — verify auto-materialized identity assets against the manifest SHA-256 pins.** `DependencySpec` carries an optional `PinnedDigest`; `ensure_downloaded` verifies after a completed download (inside the shared task so joiners share one verdict) AND for a pre-existing unattested copy found by the cache lookup. Admission deletes + refuses on mismatch with the existing typed `Sha256Mismatch` error naming the file, both digests, and `mold pull pulid-flux`; placement preview stays read-only and reports an unproven copy as pending. Identity assets require a pin; encoder ladders pass `None` (their variant tables carry no digests — separate change).
- **P2 — immutable license pin.** InsightFace terms are pinned to `raw.githubusercontent.com/deepinsight/insightface/7fadd420…/README.md` (same digest `84606d9a…baf22`), acceptance binds `(url, sha256)`, and a structural test rejects any `master`/`main`/`HEAD` license URL.

Gate: fmt, workspace clippy, core/server suites green (pre-existing env/parallelism flakes unchanged). Refs #1220.


jamesbrink:
Superseded by https://github.com/utensils/mold/pull/1242 — #1239 was squash-merged so this branch's merge commits diverged from main.

Issue #1318, opened 2026-08-23
feat(h3): execute the NVFP4 compact transformer layout


jamesbrink:
## What

Teach the compact H3 DiT loader to execute the NVFP4 layout. The artifact itself is published and installable by then; this issue is only the runtime.

**Blocked on:**
- the downloadable-variant issue (the artifact must be installable and pinned)
- the layout probe's verdict (scale layout, divergence, and measured cost)
- a provenance decision, if the probe says ship it

## Scope

**candle contract** (`crates/mold-candle/src/minimax_h3/`):
- `H3ComfyPrunedFormat::Nvfp4` + `stable_id` (`comfy_dit.rs:93-108`), and the two `H3ComfyPublishedArtifact` variants with their `file_bytes` / `content_sha256` (`:110-181`)
- the **19-byte** `{"format": "nvfp4"}` marker constant beside `PUBLISHED_INT8_COMFY_QUANT_PAYLOAD` (`:73-76`), plus its count and payload check (`:1088-1154`) and schema injection (`:2226-2249`). Note all four marker strings in the tree are distinct: DiT int8 72 B, DiT nvfp4 19 B, Qwen int8 29 B, Qwen nvfp4 54 B. `comfy_dit.rs` does a fixed-length memcmp, not a JSON parse — the marker is authentication, never dispatch.
- `detect_format` (`:2292-2320`), `require_supported_runtime` (`:278-306`), and the two `!= Int8ConvRot` refusals (`:606-648`)
- **`expected_schema` (`:2415-2570`)** — the largest single chunk. Today nothing can emit `weight_scale_2`, an F8_E4M3 `weight_scale`, or a `[rows, in/2]` U8 weight shape. New header pins are already known: `header_len = 116_000`, `header_sha256 = b62e4dffa5e111a50c42132e0c200fdebbd499b33ab7f81670cde0b498524354`, `tensor_count = 1132`.
- `load_block` (`:1613-1665`) and a sibling for `load_raw_int8` (`:1475-1486`, which hard-refuses any dtype but `"I8"`)

[...]

Issue #1290, opened 2026-08-22
feat(h3): add the NVFP4 compact transformer layout (20.97 GB -> 12.53 GB)


jamesbrink:
## Goal

Cut the compact H3 transformer from **20.97 GB to 12.53 GB (-40%)** by adding an NVFP4 layout, reusing the NVFP4 linear mold already ships. This is the shortest path to making H3 usable on a 24 GB card without the host-RAM wall documented in #1289.

## Why this variant, and not the others

Every published smaller H3 transformer, measured:

| Candidate | Size | Reuses | Verdict |
| --- | --- | --- | --- |
| **NVFP4 pruned** (`Abiray`) | **12.53 GB** | `H3ComfyNvfp4AwqLinear` - already shipped for the Qwen3-VL text encoder | **recommended** |
| mixed INT4/INT8 convrot (`Abiray`) | 15.90 GB | convrot path, but INT4 blocks are new | second choice, smaller win |
| GGUF Q4_K (`unsloth`) | 11.42 GB | nothing - GGUF layout is not the ConvRot/NVFP4 schema | biggest win, largest new code |
| `pruned_fp8_scaled` (Comfy-Org) | 20.96 GB | - | no size win at all |
| Qwen3-VL Q4_K_M / int4_convrot | 14.58 / 14.95 GB | - | saves ~1 GB over the current 15.69 GB; not worth a review cycle |

## The schema evidence

I diffed the NVFP4 checkpoint's safetensors header against the one mold ships today. The result is unusually clean:

```
current int8_convrot : 932 tensors
nvfp4                : 1132 tensors
shared keys          : 932   (every single current key is present)
only-in-nvfp4        : 200   (the `weight_scale_2` tensor scalars)
shared keys differing in shape : 600
```

Those 600 are exactly the 200 quantized linears x `{weight, weight_scale, comfy_quant}`. **All 332 remaining tensors - the AdaLN table, RoPE `inv_freq`, every bias, norm, and patch projection - are identical in name, shape, dtype, and even byte offset.** Same base checkpoint, requantized.

And the quantized tensors are precisely the layout `H3ComfyNvfp4AwqLinear` was written for:

| tensor | INT8 convrot (today) | NVFP4 | mold's NVFP4 linear expects |
| --- | --- | --- | --- |
| `weight` | I8 `[21504, 5376]` | U8 `[21504, 2688]` | `packed_weight` `(rows, in_features/2)` - two nibbles per byte, matches |
[...]

PR #1322, opened 2026-08-23
perf(h3): store NVFP4 block scales at source width (-4.57 GB host RAM)

jamesbrink:
Closes #1316. Second of the four issues that replaced #1290.

## What ships

H3's host-RAM admission floor drops **19,871,751,032 → 15,300,615,032 bytes** (the number quoted in #1289's refusal). Output is **bit-identical**: `H3ComfyNvfp4AwqLinear` stored the checkpoint's 1-byte F8E4M3 block scales as a `Vec<f32>` (4,571,136,000 B of pure expansion on the shipped Qwen3-VL conditioner); it now keeps the bytes and widens through a 256-entry LUT built by candle's own `to_dtype(F32)`, so no hand-rolled FP8 decoder can drift.

## Proof

- `nvfp4_dequantization_is_bit_identical_across_every_accepted_scale_pattern` — all 128 accepted E4M3 patterns (incl. `0x00`, `0x80` = −0.0, subnormals, 448.0), expectation derived independently from `to_dtype(F32)` on the natural grid × `E2M1_LUT` × `tensor_scale`, compared with `f32::to_bits()` for both `dequantize_weight` (partial trailing chunk exercised) and `forward_dequantized`.
- `f8e4m3_widening_table_is_candles_own_conversion` — all 256 entries bitwise, both NaN encodings.
- `nvfp4_host_scale_cache_is_one_byte_per_logical_source_scale`.
- Retained host bytes now **equal** `effective_parameter_bytes` (`assert_ne!` → `assert_eq!` in two places); `qwen_nvfp4.rs`'s source-encoded schema facts and the `H3PrivateRuntimeBoundRecord` observations are untouched — the proof this is a representation change, not a checkpoint change. No re-capture needed.
- One genuine test consequence fixed: `private_qwen_bridge_rejects_undercharge_and_post_plan_mismatch` perturbed downward, which is now an undercharge caught earlier; it perturbs upward and additionally asserts the new equality invariant.

`float8 = "0.7"` added to `mold-candle` — same dep, same comment as `mold-inference` already carries for `ltx2/fp8_widen.rs`; candle uses it internally but does not re-export `F8E4M3`. Already vendored; lockfile gains one line.

## Tier

[...]

PR #1326, opened 2026-08-23
perf(h3): NVFP4 layout probe + device-side dequantization (bit-identical, 3.9–7.0x)

### LINKED ISSUE #1317: perf(h3): run NVFP4 dequantization on the device instead of the host
## What

The Qwen3-VL conditioner mold ships **is** NVFP4, and it runs `forward_dequantized` (`crates/mold-candle/src/minimax_h3/text.rs:112`). That path has exactly one arm on every backend — `comfy_quant.rs:1364-1372` says so deliberately: *"There is deliberately no device dispatch here, on any backend: NVFP4 has no native kernel in mold at all."*

So `dequantize_rows` (`comfy_quant.rs:1316-1345`) unpacks ~24.4 G weights in a **single-threaded host scalar loop** — nibble shift, LUT index, two f32 multiplies per weight — and then uploads dense f32, once per prompt encode, on every H3 render.

Move it to the device.

## Part 1 — the layout probe (do this first)

A new unregistered example, `crates/mold-candle/examples/h3_nvfp4_layout_probe.rs`, following `examples/h3_comfy_int8_block_qualification.rs`. There is no `[[example]]` section in `Cargo.toml` (autodiscovery only), so it reaches no release surface. It must not call `open_h3_comfy_published_int8_checkpoint`; it does its own bounded header read (`serde_json` is already a dependency, `safetensors`/`memmap` are not).

It settles three things in one JSON artifact:

**(a) Are comfy-kitchen NVFP4 block scales swizzled?** `unswizzle_nvfp4_scales` (`comfy_quant.rs:939-985`) unswizzles unconditionally, and LTX-2 (`mold-inference/src/ltx2/model/video_transformer.rs:826`) and Flux.2 (`flux2/transformer.rs:329`) make the same assumption — **none of it numerically confirmed on H3-shaped tensors.** It cannot be read off a header: every H3 linear has `out_features % 128 == 0` and `blocks_per_row % 4 == 0`, so the swizzled and natural shapes coincide with zero padding.

Decide it numerically. `H3ComfyInt8ConvRotLinear::dequantize_weight` reverses the Hadamard and yields the dense logical weight, and both checkpoints are requantizations of the *same* pruned BF16 base (332 tensors byte-identical), so the INT8 dense weight is a ~0.5%-accurate oracle. Correct hypothesis => relative Frobenius error 3-6%. Wrong => ~141%, because the 128x4 swizzle permutes scales across a 512-element tile. No ambiguous middle.

**(b) Divergence**, in activation space as well as weight space, at `REVIEWED_MAX_TARGET_VIDEO_ROWS = 37_296` (`private_server.rs:4853`), with `pruned_bf16` as a common reference so `nvfp4_vs_bf16` and `int8_vs_bf16` are directly comparable.

**(c) Cost**: host dequant vs H2D vs matmul, against the native cuBLASLt INT8 arm, extrapolated to `per_linear_ms x 4 x 50 x steps`.

## Part 2 — the device arm

Candle's CUDA backend accepts `index_select` ids as **U8 directly** (`candle-core/src/cuda_backend/mod.rs:459`, `"is_u8"`), so no new kernel is needed:

1. Two 256-entry F32 LUTs for the high and low nibble, reproducing `comfy_quant.rs:1332-1336`'s exact ordering; `index_select` -> `cat` -> `reshape`.
2. One 256-entry E4M3->F32 LUT (`mold-inference/src/nvfp4.rs:168` is already the spec and is unit-tested at `:345`) -> `broadcast_mul`.
3. `* tensor_scale`, `to_dtype(BF16)`.

The CPU-storage guard (`comfy_quant.rs:1166-1174`) constrains the *stored* tensors and is not violated by a per-chunk `.narrow().to_device()`, but the load-bearing prose at `:1367-1372` must be rewritten with the equivalence test that justifies the change.

**Gate: bit-equivalence** against the CPU loop.

## Why this matters beyond the conditioner

With the device arm, a future NVFP4 transformer's per-step traffic would be 10.84 GB — *below* today's INT8 19.27 GB — which is the only condition under which an NVFP4 DiT is worth having. Without it, an NVFP4 DiT is ~19.3 G host scalar iterations and 77 GB of PCIe per denoise step.

Prefers the byte-width scale cache to land first (it wants to upload scale *bytes*), but works either way.

Estimate: 1-2 days probe + 2-3 days arm.

jamesbrink:
Closes #1317. Third of the four issues that replaced #1290.

## Part 1 — the layout probe (`crates/mold-candle/examples/h3_nvfp4_layout_probe.rs`, unregistered)

Record: `docs/qualification/minimax-h3-nvfp4-layout-probe.{md,json}` — one real run on hal9000 (RTX 4090) against Comfy-Org's SHA-verified `pruned_bf16` as the common base (332/332 untouched tensors byte-identical across all three files).

- **Scales are swizzled** (0.095 vs 0.570 rel-Frobenius, 12/12 linears, no overlap). `unswizzle_nvfp4_scales` is right on H3 shapes — and so is `mold_inference::nvfp4`'s unconditional unswizzle that LTX-2 and Flux.2 rely on, now numerically confirmed on a real comfy-kitchen checkpoint.
- **The third-party NVFP4 DiT is a no-go as-is**: 6.4–7.6× INT8 ConvRot's activation-space error against BF16, failing #1317's own 2× rule on every linear (recorded on #1318).
- **Cost**: host dequant + H2D 237 ms vs 41 ms on the device at 37,296 rows.
- **Found**: candle's CUDA `index_select` reserves `u8::MAX` as a padding sentinel, so a U8-id LUT silently zeroes every `0xff` payload byte (122,820 in the probed qkv weight).

## Part 2 — the device arm

`H3ComfyNvfp4AwqLinear::forward_dequantized` gains a CUDA arm (`select_h3_nvfp4_linear_kind`, a pure device function mirroring the INT8 selector): packed U8 + unswizzled scale bytes are uploaded as stored and widened on the device via `index_select` lookup tables with **U32 ids** (the U8 + `0xff`-repair alternative was measured, slightly slower), then the same full-precision matmul as before — a cost choice, never a numerical one. Storage stays CPU-resident. Metal deliberately stays on the portable arm: the equivalence rests on a bit-for-bit CPU↔device test with no runnable Metal twin.

- CPU↔CUDA **bit-equality** of the dequantized weight over packed bytes `{0x00, 0xff, 0x0f, 0xf0, …}` and every E4M3 scale class, with a partial trailing chunk; CPU-only LUT index-math test for non-CUDA CI.
- `device_weight_staging_bytes` bounds the per-chunk transient device peak; the Qwen conditioner's memory authority charges the larger arm so admission cannot under-price the device path.
- Doc prose that claimed "no device dispatch on any backend" rewritten with the sentinel fact.

**Measured** (4,096 rows, 4090, production model co-resident): `attn.qkv_proj` 141.7 → **36.5 ms** (3.9×); Qwen `down_proj` shape `[5120,25600]` 286.7 → **40.8 ms** (7.0×). Every H3 render runs the Qwen3-VL conditioner through this path.

## Gates

fmt · clippy `-D warnings` (workspace, and `mold-ai-candle --features cuda`) · `mold-ai-candle` 262 CPU / 275 CUDA tests · example tests 8/8 · changelog contract PASS. Benchmark re-run independently by the orchestrator.

https://claude.ai/code/session_011D2J2pGNeyzEaS6cka2H8o


jamesbrink:
Implementation notes from the Part-2 agent, verified by the orchestrator (benchmark re-run independently):

- **Id width — U32, on two grounds.** The U8 + `packed == 0xff` repair is not bit-exact *as an addition*: `index_select` zeroes the sentinel, so every non-sentinel element gets `+0.0`, and E2M1 entry 8 **is** `−0.0` — the repair flips the sign of every negative zero in the weight, invisible to any tolerance test (pinned by `nvfp4_u8_id_repair_by_addition_loses_negative_zero`). A correct `where_cond` repair measures 6.8% slower than the U32 cast. The CUDA bit-equality test was falsified by forcing `DType::U8` ("6 of 200 elements differ in bits") after the fixture was fixed to put `0xff` in live rows of both the first and the trailing chunk.
- **What is bit-identical vs toleranced.** The *dequantized weight* is compared bit-for-bit CPU↔CUDA (chunks 1/2/5/64, out=5/in=40 so both the trailing chunk and padded narrow run). `forward_dequantized` CPU vs CUDA is ≤ 2e-4 — pure cuBLAS accumulation order on an unchanged-dtype GEMM.
- **Memory accounting — no pinned number moved.** `portable_weight_staging_bytes` had no non-test consumer, and `H3QwenNvfp4RuntimeMemoryFacts` is hashed into the authenticated-authority identity, so no field was added there. Instead `released_h3_qwen_nvfp4_max_weight_staging_bytes()` walks the released schema's packed shapes and takes the max over both arms: `mlp.down_proj [5120,25600]` → **98,716,676 B**, 2.37% of the measured 4,168,069,120 B `qwen_activation_workspace_bytes` grant that already bounds QwenEncode. The staging formula is the sum of every transient (~15.06·P), deliberately pessimistic. `H3PrivateRuntimeBoundRecord` untouched.
- **Why the second benchmark shape.** `mlp.down_proj` is the widest released conditioner projection and 6,065 rows (2,033 text + 4,032 vision) is what an FL2VA render actually encodes; 37,296 is the DiT's row count, where the 9× larger GEMM hides most of the host loop.
- Implementation trap recorded in code: candle's `to_device` copies the whole underlying storage of a `narrow()`ed view, so the chunk's bytes are taken host-side via the layout-respecting `to_vec1` and exactly those are uploaded.
- Metal stays on the portable arm until a `cfg(feature = "metal")` twin of the bit-equality gate can run on Apple hardware.

`mold-ai-inference --lib`: 2115/2 with a byte-identical failure list to pristine `main` (the two `pickle_convert` timestamp tests).

jamesbrink:
Codex peer review (`codex review --base origin/main`): no findings — *"The CUDA and CPU/H3 targets compile successfully, and the relevant quantization tests pass. No discrete correctness, security, or maintainability regression was identified in the changed execution or accounting paths."*

jamesbrink:
End-to-end on hal9000 (see #1322's acceptance comment): the `0.24.0` build with this device-dequant arm produced frames and audio **byte-identical** (`b543bbf5…` / `ea0438be…`) to a build using the host loop, on the full 124-frame Turbo 8-step render.

PR #1379, opened 2026-08-25
fix(windows): unblock Authenticode verification in CI

jamesbrink:
No Windows installer has ever been produced, on any branch, since the Windows
target landed. Every run reached the signing step and then stopped — including
the `v0.24.0` and `v0.25.0` release tags, which is why both shipped macOS only.

## What was wrong

`scripts/verify-windows-signatures.ps1` imported the signing certificate into
`Cert:\CurrentUser\Root` so `Get-AuthenticodeSignature` would report `Valid`
rather than the `UnknownError` an untrusted self-signed chain always produces.

**Writing the user Root store requires interactive trust confirmation.** Under
`-NonInteractive` it fails outright:

```
Import-Certificate: verify-windows-signatures.ps1:16
     | UI is not allowed in this operation.
```

But GitHub's own shell, from the job log, is:

```
shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
```

No `-NonInteractive`. So the call did not fail — CryptoAPI raised a modal trust
dialog on a session desktop nobody could see, and the step blocked until the
job's six-hour ceiling.

## Blast radius

| Where | What happened |
| --- | --- |
| `desktop.yml` on `main` | ~30 runs checked; **zero** produced `mold-desktop-windows-x64-self-signed`. Four were wedged on this step simultaneously, the oldest for 3h. |
| `release.yml` tag `v0.25.0` | `build-windows` sat in "Verify and package Windows artifacts" **13:02 → 18:04**, then cancelled. |
| …consequently | `publish` failed; `release-native`, `release-latest`, `release-containers` and both AUR jobs **skipped**. |
| `v0.24.0` | Same shape. |

Because `publish` needs `build-windows`, one line stalled the Linux tarballs,
the container matrix, and AUR for two consecutive releases — not just Windows.

`scripts/tests/windows-self-signing.sh` stayed green throughout because it
greps the verifier for `Get-AuthenticodeSignature` and never runs it.

## The fix

Trust was never what this check is for; it exists to prove the artifacts carry
an intact signature made by the expected certificate. So:

- **Pin the thumbprint** against `$signature.SignerCertificate`, and against
  the supplied `.cer` itself, instead of trusting the chain.
- **Accept `UnknownError` alongside `Valid`** — that is precisely the status an
  untrusted self-signed chain yields. Every other status stays fatal, so
  `HashMismatch` and `NotSigned` are still refused.
- **Keep chain validation** via `X509Chain` with
  `AllowUnknownCertificateAuthority`, so an expired or malformed signer still
  fails. Only the untrusted root is waived.
- **No certificate store is written at all**, so the script is safe under any
  invocation.

The contract test now also asserts the verifier writes no certificate store.
Verified it **fails** against the old script before the fix was applied.

## Verification

Windows 11 ARM64 (Snapdragon Surface). Every case run through GitHub's exact
invocation — `pwsh -command ". '<generated>.ps1'"`, no `-NonInteractive` — in a
child process with a timeout, so a block is observable rather than hanging the
harness.

Against the **real release build**, signed the way CI signs it (sha256 +
DigiCert timestamp):

| Artifact | Old script | This PR |
| --- | --- | --- |
| `mold-desktop.exe` (77.9 MB) | blocked past 60s | **verified, 8.5s** |
| `Mold_0.25.0_arm64-setup.exe` (18.8 MB) | blocked past 60s | **verified, 8.5s** |

Refusals, all still correct:

| Case | Result |
| --- | --- |
| tampered byte in a signed image | refused — `HashMismatch` |
| signed by a different certificate | refused — unexpected certificate |
| unsigned binary | refused — `NotSigned` |
| `.cer` that is not the pinned certificate | refused |
| malformed thumbprint argument | refused |
| missing file | refused |

Also green: `windows-self-signing.sh`, `windows-helper-shells.sh` (both
PowerShell editions agree), `shellcheck`, `prettier`.

**App UAT** — built and launched from this branch after the final rebase:

- `scripts\windows.ps1 build` → exit 0, **6m56s**, feature recipe `pulid`
- window opens titled `Mold`, stays up, 55 MB / 43 threads
- embedded engine listening, `/api/status` → `401` unauthenticated, and with
  the app's own key `/api/status`, `/api/models` (621 KB), `/api/capabilities`
  all → `200`

## Second commit: the desktop lockfile

`desktop/src-tauri/Cargo.lock` was out of sync with its manifests — recent
server work added `chacha20poly1305` and its tree, and the desktop cargo root
is `exclude`d from the workspace so a workspace-wide refresh never reaches it.
`cargo metadata --locked` **failed** on the committed file; any build silently
rewrote it. Regenerated by building; no manifest edited.

## Notes for review

- The behaviour matrix above is a local harness, not committed. Committing it
  would mean creating and trusting certificates inside CI, which is the class
  of operation that caused this outage.
- One unrelated flake seen once in ~30 runs: `Install protoc and nasm` failed a
  download. Not this defect, not addressed here.
- I have **not** run `scripts\windows.ps1 test` on this tree; it was stopped
  mid-compile earlier and never produced results. This PR touches one
  PowerShell script and one shell test, neither reachable from that suite.

### DIFF
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -59,6 +59,7 @@ on:
       - "scripts/verify-windows-signatures.ps1"
       - "scripts/tests/windows-helper-shells.sh"
       - "scripts/tests/windows-self-signing.sh"
+      - "scripts/tests/windows-signature-verifier.ps1"
       - "scripts/prune-desktop-nightly-assets.sh"
       - "scripts/select-desktop-stable-latest.sh"
       - "scripts/tests/check-desktop-nightly-main-head.sh"
@@ -157,6 +158,7 @@ jobs:
               - 'scripts/verify-windows-signatures.ps1'
               - 'scripts/tests/windows-helper-shells.sh'
               - 'scripts/tests/windows-self-signing.sh'
+              - 'scripts/tests/windows-signature-verifier.ps1'
               - '.github/workflows/desktop.yml'
               - '.github/workflows/desktop-distribution.yml'
             rust_format:
@@ -287,6 +289,10 @@ jobs:
         run: |
           bash ../scripts/tests/windows-helper-shells.sh
           bash ../scripts/tests/windows-self-signing.sh
+      - name: Exercise Authenticode verification with an ephemeral signer
+        shell: pwsh
+        timeout-minutes: 2
+        run: ../scripts/tests/windows-signature-verifier.ps1
       - run: bun install --frozen-lockfile
         working-directory: .
       - run: cargo fmt --manifest-path src-tauri/Cargo.toml -- --check
--- a/changelog.d/windows-signature-verification-hang.md
+++ b/changelog.d/windows-signature-verification-hang.md
@@ -1,0 +1,10 @@
+- **Windows release artifacts publish again.** Authenticode verification no
+  longer imports the self-signed certificate into the user Root store to force
+  a `Valid` status. That import needs interactive trust confirmation, and CI
+  runs PowerShell without `-NonInteractive`, so it blocked on a dialog nobody
+  could see until the job's six-hour ceiling — stalling every Windows installer
+  and, because publication waits on the Windows build, the Linux, container,
+  and AUR assets with them. The check now pins the signing thumbprint against
+  each signature, so tampered, unsigned, and wrongly signed artifacts are still
+  refused
+  ([#1379](https://github.com/utensils/mold/pull/1379)).
--- a/desktop/src-tauri/Cargo.lock
+++ b/desktop/src-tauri/Cargo.lock
@@ -9,6 +9,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
+name = "aead"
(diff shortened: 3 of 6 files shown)


jamesbrink:
Implemented the peer-review follow-ups in `f0343554`.

- Restricted PowerShell's broad `UnknownError` fallback to the exact localized `CERT_E_UNTRUSTEDROOT` message (`0x800B0109`). Timestamp, provider, and other unmapped WinVerifyTrust failures remain fatal.
- Added a Windows integration test that creates an ephemeral code-signing certificate in `CurrentUser\\My`, verifies a valid signed fixture, and rejects tampered, unsigned, and wrong-signer fixtures.
- Added a two-minute timeout so this regression cannot silently consume another six-hour job.
- Wired the test into the Windows PR job and both workflow path classifiers.

Local validation passed: Windows signing contract, locked desktop Cargo metadata, PowerShell parsing, workflow validation, and `git diff --check`.

Independent subagent peer review found no blocking issues. It confirmed the HRESULT match, certificate cleanup, workflow paths, negative cases, and timeout. Its only non-blocking P3 note is that the integration fixture exercises the PowerShell Authenticode SIP rather than a PE/NSIS executable; the corrected WinVerifyTrust trust-error path is shared.

PR #1397, opened 2026-08-26
fix(queue): unify durable generation admission

jamesbrink:
## Summary

- replace family-specific placement and admission paths with one queue-first durable generation contract across images and video
- persist accepted, held, failed, retryable, cancelled, and restart-recovered work in SQLite with idempotent client reconciliation
- harden authority fencing, concurrent admission, encrypted media handling, and explicit error reporting so submissions cannot silently disappear
- migrate desktop, web, iPhone, Android shared UI, CLI, TUI, MCP, RunPod, documentation, and the shipped Mold skill to the unversioned canonical routes
- restore explicit missing-model confirmation and tracked pull progress across desktop, web, and mobile without reintroducing blocking placement preview
- remove superseded multi-output routing instead of retaining legacy endpoints

## Reliability behavior

- admission returns after durable queue commit instead of waiting for model placement or load
- pinned and automatic routing preflight cached inventories; absent models require confirmation, remain visibly tracked while pulling, and resume on the frozen target
- queued cancellation immediately revokes claimed preparation, releases queue capacity, and prevents later worker dispatch
- cancelled and failed work remains persistently visible with its terminal state until explicitly removed; restart recovery preserves batch membership and retry effects
- unavailable work remains visible and retryable with its error
- server restarts recover queued and interrupted work without duplicate execution
- direct singleton compatibility remains bounded while durable batches use one shared orchestration layer
- maintenance and unserviceable requests fail before persistence

## Verification

- independent final peer review: clean, with no guide or documentation files added
- Rust 1.93 MSRV workspace and feature checks
- workspace Clippy with warnings denied; Rustfmt and generated-profile checks
- affected mold-ai-db queue suite: 289 passed
- affected cancellation, preparation, capacity-release, pause/resume, and shutdown-retention tests pass
- final affected desktop/mobile suite: 437 passed
- desktop production typecheck and build
- full desktop suite: 5147 passed
- full web suite: 1742 passed; production build passed
- full shared Studio suite: 1291 passed; architecture check passed
- current-head GitHub CI pending

Draft only. Do not merge or deploy.

---

## Review remediation (2026-08-26)

Two independent review passes plus adversarial planning found that four defects were one drift: `capabilities.queue.heterogeneous_batch`, `capabilities.durable_media`, `/api/status`, `/health`, `POST /api/generation-batches` and direct admission each re-derived "is durable admission available" from a **different subset of the same four conjuncts**, and the gaps between those subsets were the defects. Three tests were red on this branch — two the PR added itself, asserting a contract the code did not implement.

### Fixed

- **`routes::DurableAdmissionReadiness` is now the single authority.** Total over all four conjuncts at one evaluation point; takes `config` as a required parameter it never reads, so a caller without a config cannot call it at all. `DurableAdmissionUnready` retains *which* conjunct failed, because the two refusal codes and `durable_media_status().reasons` both report a per-conjunct cause.
- **P0 — a degraded encrypted-media store no longer breaks all conditioned generation.** The media gate ignored `explicitly_requested` while the gate six lines below it consulted it, so on `MOLD_DB_DISABLE`, an unresolvable `MOLD_HOME`, widened store permissions, or a corrupt admission key, **every** `source_image` / `id_image` / `edit_images` / `mask_image` / `control_image` / `audio_file` / `source_video` / `extend_video` / keyframe / H3 request returned 503 on both direct routes — behind a status both client classifiers read as transient, on a host serving `/health` 200. It now falls back to the attached path; an explicit `X-Mold-Operation-Id` still gets the typed refusal.
- **P0 — the capability no longer advertises a protocol the same server refuses.** A host whose `generation-admission.key` is corrupt or missing boots and serves, and used to promise `admission_protocol_version: 2` and then 503 every batch POST with no client fallback.
- **P0 — `capabilities.durable_media` regained its applicability gate**, so it can no longer contradict `/api/status` and `/health`.
- **P0 — durable replay asks the same MiniMax H3 identity question admission asked**, so a download-only identity answers `MINIMAX_H3_RUNTIME_UNAVAILABLE` instead of parking as a permanent hold.
- **`mold run` and the TUI keep singletons on the attached path**, restoring live step progress, the centered fixed-protocol denoise preview that CLAUDE.md protects by name, and remote missing-model auto-pull. Batch N still uses durable batch admission; durability is unaffected because `/api/generate` admits through the same gate either way.
- A direct one-shot no longer carries synthesised prepared-batch provenance.
- An inference panic on the single-worker path settles non-retryable instead of offering a re-dispatch of work that just panicked the process.
- `retry_held_owned` no longer clears `replay_seen`, the only bound on a boot crash loop.
- `/api/generation-batches/status` is rate-limited as a read; `POST /api/queue/{id}/retry` with generation.
- Client reconciliation bounds transient-failure retries while leaving the wait for queued work unbounded.

### Deliberate behaviour decisions

- **A degraded media store now downgrades instead of refusing.** This trades a hard 503 for a silent loss of the durability guarantee on that request. It is the right trade — the refusal broke all conditioned generation behind a transient-looking status nothing fell back from — but it is a decision, not a refactor.
- **Direct durable admission deliberately does not require `v2_authoritative`.** The durable feeder is spawned on `start_generation_runner` alone and runs in legacy and observe dispatch, so a direct generation there is genuinely durable. An earlier draft added the conjunct "for consistency" and would have deleted real durability from every legacy and observe host; this branch's own `headerless_direct_media_is_durable_without_a_client_operation_id` proved the omission was intentional. The batch route keeps the V2 requirement, because that is what the capability advertises.
- The two routes keep **opposite refusal precedence** — direct reports gallery output first, batch reports scheduler/journal/admission first — and a test pins both.

### CI

`rust` was failing on this branch with 5 test failures. Three were the real defects above. The other two (`a_cpu_rendered_generation_stamps_its_job_id_and_clears_its_row`, `direct_admission_rebinds_a_stable_gpu_after_restart_and_renumbering`) passed locally and failed only under CI's parallel load; the first sampled a durable row that clears asynchronously and now waits for it with the same bounded-wait shape its sibling test uses.

Local: clippy `-D warnings` clean, rustfmt clean, mold-ai-server 1999 passed, mold-ai-db 284, mold-ai-core 1506, mold-ai 684, mold-ai-tui 769, vitest studio 1291 / web 1742 / desktop 5150.

### Added after the adversarial-planning synthesis

- **`POST /api/queue/{id}/retry` is the first transition that moves a job backward through the browser reducer's `FORWARD_PHASE_RANK`** (held 2 -> queued 1). That reducer accepts a backward move only when the snapshot is strictly newer, and `revision` is never supplied on the wire, so it falls through to `updated_at_ms` -- which `retry_held_owned` wrote with the same `now_ms` the preceding hold could carry. A retry inside one millisecond of its own hold was silently dropped and the client kept rendering Held for a job the server had queued. Both UPDATEs now write `MAX(?, updated_at + 1)`; fixing it server-side corrects every client at once and leaves the reducer's tie-break alone.
- `FailureDisposition` and `PreparationDisposition` were byte-identical three-variant enums in one module, one of them `#[allow(dead_code)]`. Collapsed.
- **A5 (capability-derived `references` refusal) was considered and rejected as fail-open.** `queue_media_admission.rs:826` refuses ordered references because the durable protocol cannot persist their staging authority -- a structural fact, not a capability toggle. Deriving the refusal from `advertised.h3_references` would mean that flipping that constant admits requests whose authority the server cannot persist.

### Follow-ups, now closed

All four items previously listed here as known-but-unaddressed are fixed on this branch. Two of them were one missing primitive.

**A durable batch child carries a monotonic `revision` (schema v29).** It is incremented by every authoritative state transition and by nothing else — a *refused* transition must not spend one, or a reconciling client cannot tell it from one that landed. `updated_at_ms` had been doing two jobs at once: an honest wall clock, and the version every client reducer compares. It is the weaker of the two — SQLite commits several transitions inside one millisecond routinely, and `POST /api/queue/{id}/retry` is the only route that moves a child *backward* through `FORWARD_PHASE_RANK` (held → accepted), so a collision there is not a tie to break arbitrarily: it decides whether the retry is visible at all. `childRevision` reads the field off the child rather than from an action parameter nobody ever supplied, and reads both absence and `0` as *no authority*, because rows admitted before the migration sit at `0` until their next transition. The field is additive, so an older server deserializes to `0` and every client falls back to the timestamp comparison; the retry path keeps its `MAX(?, updated_at + 1)` bump for exactly those clients.

**The same primitive fences the ambiguous retry.** A retry POST whose response was lost could not be reconciled from state alone: `held` conflates "the retry landed and the job was immediately re-held for a new reason" with "the retry never arrived". On the last bounded attempt the old code published the pre-retry snapshot as fresh authority while the original POST might still commit. It now compares against the revision captured *before* the POST and **retains** the fence when the retry did not land. A host with no revision authority keeps the original publish-on-exhaustion behaviour — with no version to compare there is no evidence either way, and hanging forever is the worse failure. MCP gets the same rule: an advanced revision settles on the first read instead of burning five polls, and an unadvanced one composes the un-landed retry onto the settled error.

**Held rows have a retention sweep.** `queue.held_retention_days` (default 30, `0` keeps forever, env `MOLD_QUEUE_HELD_RETENTION_DAYS`) purges abandoned holds hourly and through `POST /api/queue/held/sweep`. A hold is durable so a human can return to it; one nobody returns to kept a `GET /api/queue` row forever **and** pinned its encrypted media, because `hold_claimed` retains `media_set_id` on purpose so a retry still has its source images. The sweeper never writes `generation_queue_media`: deleting the queue row fires the existing retire trigger, and it collects what that trigger marked. Age runs from when the row was *held*, a retry or cancel landing first wins, and the batch child is settled to `failed` rather than deleted — after the queue row is gone that summary is the only thing a reconnecting client can read.

**`GET /api/gallery` accepts `?filename=`.** Hydrating one artifact no longer serializes and transfers the whole gallery index. It is a filter on the existing endpoint rather than a second route on purpose: the listing's DB read, archive overlay, organization overlay and filesystem fallback are one pipeline with five exits, and a dedicated handler would restate all of it as a second authority that can disagree about what a print's metadata is.

### Peer review of the follow-up work

Two Codex passes, five findings, all real and all fixed:

1. The retry fence reached the submission but not MCP's reconciliation, which had reduced the ambiguous result to a boolean.
2. `candidate_for_job(...).ok().flatten()` turned a media-lookup failure into "no media" — the row was purged, its bytes left `gc_pending`, and `media_deferred` still reported zero.
3. `migrate_config_toml_to_db` never imported a hand-written `[queue]` section, and the migration rewrites `config.toml` as bootstrap-only, so the configured value was lost rather than left in the file.
4. The sweep ran its SQLite and filesystem work on the async task instead of the blocking pool.
5. `POST /api/queue/held/sweep` was on the router with a `#[utoipa::path]` annotation but absent from `ApiDoc`'s paths list, so `/api/openapi.json` omitted it.

Local after the fixes: clippy `-D warnings` clean, rustfmt clean, mold-ai-server 2006, mold-ai-db 293, mold-ai-core 1510, mold-ai 687, mold-ai-tui 770, vitest studio 1297 / web 1744 / desktop 5161.

Three `mold-ai-inference` tests (`pickle_convert` ×2, `minimax_h3::vae_runtime` ×1) fail non-deterministically on this ext4 box through inode reuse plus timestamp granularity, on files this branch does not touch. Same documented environmental family as `download::tests::durable_attestation_...`; not a regression here.


jamesbrink:
Closing while the branch keeps moving: all legacy/compat generation paths are being removed and the queue stall on hal9000 fixed on `fix/unified-generation-admission`. A fresh PR will be opened when the branch is ready for review.

PR #1411, opened 2026-08-26
fix(studio): repair Ref2VA reference submission

jamesbrink:
## Summary
- compute and bind SHA-256 provenance when Ref2VA references are attached from local files or Library prints
- keep auth-disabled hosts on the validated inline Ref2VA path while authenticated hosts use request-bound uploads
- advertise explicit authless-inline capability so stale keys work and unknown legacy capability snapshots remain fail-closed
- preserve API-key protection for upload handles, server paths, descriptor/mixed reference sets, durable batches, and chains

## Verification
- `nix develop -c cargo check --workspace --all-targets`
- `nix develop -c desktop-check`
- Studio Vitest: 113 files, 1,274 tests passed
- web Vitest: 106 files, 1,739 tests passed
- desktop Vitest: 400 files, 5,112 tests passed
- focused H3 Metal server auth/capability regressions passed
- changelog, Rustfmt, Prettier, and diff checks passed

## Review
- independent agent peer review completed through three security edge cases; all findings addressed; final exact-diff review reported no findings

### DIFF
--- a/changelog.d/ref2va-library-reference-hash.md
+++ b/changelog.d/ref2va-library-reference-hash.md
@@ -1,4 +1,6 @@
 - **MiniMax H3 Ref2VA Library references now queue correctly.** Mold binds each
   selected image to its content hash before placement, so Library and local-file
   references work across web, desktop, iPhone, and Android instead of being
-  refused with a missing SHA-256 error.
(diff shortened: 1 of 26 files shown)


jamesbrink:
Superseded by the clean replacement branch with the identical reviewed tree and no empty retrigger commit.

PR #1440, opened 2026-08-28
perf(desktop): Library store/render hot paths + regression guards (gallery perf 1/4)

jamesbrink:
## Summary

First of four serialized PRs making the desktop Library near-native at 1 000+ prints (plan: store/render → persistent thumbnail cache → server thumbnails → web/mobile parity).

- **Store**: `organizationIndex` / `bucketIndex` getters compute each logical print's organization union and its cross-host copies ONCE per data change. Before: `unionOrganization` re-ran per call across ~10 full-N getters (28 200 calls for 3 000 prints on one change) and every visible tile scanned every bucket (25.2 M identity comparisons to resolve copies at 3 000 prints).
- **Grid**: per-tile view models computed once; ONE flat print-keyed absolutely positioned tile layer so a thumbnail-size drag or resize moves tiles instead of remounting `AuthedMedia` (which cancelled and refetched the thumbnail); memoized model labels; settled slider persistence; `contain` on tiles, transform-only hover transition.
- **Rows are raw immutable snapshots** (`markRaw`, replaced via `patchRow`) — no deep proxies over 1 000 × 60-field metadata objects.
- **304 poll is a no-op**: `fetchBucket` skips the assignment when the conditional listing hands back the same array.
- **Scheduler**: per-(priority, host) FIFO queues, O(hosts) dispatch (was copy+sort of the whole queue per dispatch); `AuthedMedia` takes a `priority` prop (visible / near from the virtual window) and `decoding="async"`.
- **History ▸ Runs** capped like Sequences.

## Regression guards (operation counts, not wall clock)

- `desktop/src/stores/gallery.perf.test.ts` — 3 000 prints across 2 hosts with mirrored and legacy copies: ≤ 1 union pass per logical print across every getter, 0 union passes on a filter change, 0 identity scans to resolve every print's copies.
- `desktop/src/views/LibraryView.perf.test.ts` — mounts the REAL virtualizer + justified layout at 2 000 prints: ≤ 80 tiles in the DOM, 0 `AuthedMedia` remounts across a three-tick slider drag.
- `studio/lib/thumbnailScheduler.test.ts` — drains 10 000 tiles with zero `Array.prototype.sort` calls and ≤ 2Q dispatch scans.
- `studio/lib/galleryPerfBudget.ts` — the shared `expectOpsUnder` helper every guard reads.

## UAT

Dev app against `MOLD_HOME=/Volumes/ExternalStorage/mold2` with plato + hal9000 connected: 1 206 prints in the All view render and scroll smoothly.

## Test plan

- [x] `cd desktop && bun run test` (5 342 passed)
- [x] `bun run test:studio`, `bun run check:architecture`, `bun run check:dead-code`
- [x] `vue-tsc -b`

### DIFF
--- a/changelog.d/gallery-perf-desktop-store.md
+++ b/changelog.d/gallery-perf-desktop-store.md
@@ -1,0 +1,12 @@
+- **Desktop Library stays smooth past 1 000 prints.** The gallery store now
+  indexes each print's organization and cross-host copies once per data change
+  instead of re-deriving them for every tile on every render (one filter change
+  used to run ~10 full passes over the library, and each visible tile scanned
+  every bucket), the grid renders one flat print-keyed tile layer so a
+  thumbnail-size drag or window resize moves tiles instead of remounting their
+  media, a 304 poll no longer invalidates the merged grid, gallery rows are
+  raw immutable snapshots rather than deep-reactive proxies, the thumbnail
+  scheduler dispatches in O(hosts) instead of re-sorting the whole queue, and
+  History ▸ Runs is capped like Sequences. Operation-count regression guards
+  (`gallery.perf.test.ts`, `LibraryView.perf.test.ts`, the scheduler drain
+  test) fail CI if any of these hot paths regress.
--- a/desktop/src/components/gallery/AuthedMedia.vue
+++ b/desktop/src/components/gallery/AuthedMedia.vue
@@ -1,6 +1,10 @@
 <script setup lang="ts">
 import { onMounted, onUnmounted, ref, watch } from "vue";
-import { galleryThumbnailScheduler, type ThumbnailHandle } from "@studio/lib/thumbnailScheduler";
+import {
+  galleryThumbnailScheduler,
+  type ThumbnailHandle,
+  type ThumbnailPriority,
+} from "@studio/lib/thumbnailScheduler";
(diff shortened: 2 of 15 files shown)


jamesbrink:
Codex peer review (gpt-5.6-sol, high reasoning) findings and how ae973528e addresses them:

1. **Scheduler host starvation** — `nextRunnable` never rotated hosts, so two busy hosts could refill freed slots ahead of a third forever. Fixed: the winning host moves to the back of its priority map (round-robin); new fairness test `rotates hosts so a third host is not starved by two busy ones`.
2. **Indexed copy lookup changed observable order** — the old scan returned matches in bucket order; the index emitted the exact filename first. Fixed: `BucketIndex` keeps row ordinals and `locationsIn` sorts matches by ordinal; new test `returns a print's copies in bucket order, legacy twin before canonical row` (cover selection reads the first copy per host).
3. **Guards could pass after a scan returned** — counting `sameLogicalGalleryPrint` misses an inline `.find()`; the scheduler trusted its own counter. Fixed: the gallery guard wraps every bucket array in a counting `Proxy` and budgets raw element reads; the scheduler guard counts `Map` iterator steps externally.

No actionable regression was found in the `markRaw`/`patchRow` paths, the flat tile layer interactions, or the 304 no-op.

PR #1442, opened 2026-08-28
feat(desktop): persistent thumbnail cache + mold-thumb protocol; offline tiles never full-size (gallery perf 2/4)

jamesbrink:
## Summary

Second of four serialized Library-performance PRs (after #1440). Lightroom-style persistent thumbnails on the desktop.

- **`desktop/src-tauri/src/thumbnail_cache.rs`** — content-addressed disk cache under `<app_data>/thumbnail-cache/v1`, keyed `sha256(origin \0 filename \0 media_version \0 size)` (origin = `local` or a digest of the host's base URL — never the API key), 512 MB / 20 000 files, LRU by mtime, magic-sniffed on read, atomic temp+rename, per-digest singleflight.
- **Two halves per tile:** `prepare_gallery_thumbnail` (cache-first; a miss fetches natively from the host or, with the engine Off, renders this device's file in-process) and the async **`mold-thumb://`** protocol that only reads the cache — no bytes, blobs, or object URLs pass through JS for a tile; WebKit decodes off-thread and keeps its bitmap cache. `probe_gallery_thumbnails` (stat-only) and `forget_gallery_thumbnail` (delete forever) complete the bridge.
- **Offline "This device" tiles are real thumbnails now** — `galleryMediaPath(…, "local", thumbnail=true)` used to return the full-size `mold-local:` file (and mounted a `<video>` per clip). The offline renderer checks the server's shared `MOLD_HOME/cache/thumbnails` first, then renders through the new `mold_server::thumbnails` module.
- **Pre-warm** (`lib/gallery/thumbnailPrewarm.ts`): after a listing lands, tiles around the viewport are prepared at `near`/`background` priority through the shared scheduler, capped per host; a tile scrolling into view dedupes onto its in-flight prewarm.
- **Server module `thumbnails.rs`** (shared renderer, JPEG q85 with alpha kept as PNG) and **first-frame-only video decode** (`extract_first_frame`; the old poster path decoded the whole clip).

## Guards

- `thumbnail_cache.rs`: hit/miss/round-trip, version and tier in the key (API key not), prune by bytes and by count (LRU survivors), corrupt entry deleted+missed, stale tmp reaped, singleflight, traversal rejected.
- `gallery.rs`: `cached_thumbnail_performs_zero_fetches`, `offline_local_prefers_shared_server_cache_over_render` (never the full-size bytes), protocol URL round trip + traversal rejection.
- `media.rs`: `first_frame_extraction_stops_after_frame_zero` (≤ 8 checkpoints vs ≥ 30 for a full decode).
- JS: `prepareNativeThumbnail` tier/fallback/cancellation, `galleryMediaPath` local thumbnail never `mold-local:`, prewarm ordering/cap, existing suites updated.

## UAT

Dev app (`desktop-dev`) against `MOLD_HOME=/Volumes/ExternalStorage/mold2` with plato + hal9000: 1 206 prints in All; first Library open persisted 1 029 tiles (81 MB) and later opens paint from disk with no host requests.

## Test plan

- [x] `cd desktop && bun run test` (5 350), `vue-tsc -b`, `test:studio`, `check:architecture`, `check:dead-code`, prettier
- [x] `cargo test --manifest-path desktop/src-tauri/Cargo.toml --no-default-features --features metal,h3,pulid --lib thumbnail` + clippy `-D warnings`
- [x] `cargo test -p mold-ai-server thumbnails`, `cargo test -p mold-ai-inference --features mp4 first_frame`, clippy


jamesbrink:
Codex peer review (`codex review --base main`) findings, addressed in the follow-up commit:

1. **P1 — offline render without the lifecycle authority.** `fetch_local_thumbnail` now consults `local_gallery_authority`: a running server answers through its authenticated HTTP API; only a proven-Off lifecycle lock renders from the output dir, and the guard is held for the whole render. The `mold-thumb` cache-miss path goes through the same helper.
2. **P2 — prewarm dedupe stranding visible tiles.** The prewarm task now REJECTS when the native route refuses, so a visible `AuthedMedia` that deduped onto it falls back through its own retry (blob route) instead of settling on `null`.
3. **P2 — 512 tier mislabeled on older servers.** A 512 request whose bytes probe (PNG IHDR / JPEG SOF) as ≤ 256 px is filed under the 256 tier, the origin is marked downgraded for the session, and the 256 URL is returned. The server side of `?size`/`?fmt` lands in PR 3.
4. **P2 — eviction on every permanent delete.** `forgetThumbnailFor` keys on the same effective version the tile used (`media_version` or the `timestamp:size` fallback) and runs on direct delete forever, bulk delete forever, and Empty trash.

Also: the CI `Native Rust` failure was rustfmt on the excluded desktop crate (`cargo fmt --all` at the workspace root does not cover it) — formatted.

PR #1443, opened 2026-08-28
feat(server): thumbnail renditions, parallel warmup with orphan sweep, listing recency index (gallery perf 3/4)

jamesbrink:
## Summary

Third of four serialized Library-performance PRs (after #1440, #1442). Server-side thumbnail work every surface benefits from.

- **`GET /api/gallery/thumbnail/:name?size=256|512&fmt=png|jpeg`** — a rendition per request (retina tiles as 512 px JPEG q85, ~¼ the bytes of PNG; transparent prints stay PNG). The default 256 px PNG keeps its exact cache path and ETag, so older clients, the TUI, and the desktop's shared-cache lookup are byte-for-byte unchanged. Content type is sniffed from the bytes, never the extension. Unknown values are a 422.
- **Warmup** keeps its per-entry publication-gate observation contract (the existing test still pins it) but collects misses newest-first and renders them on a bounded `rayon` pool (≤ 4 threads) instead of one core serially; each render job holds its own read guard.
- **`thumbnails::sweep_orphans`** after warmup deletes versioned tiles no live or `.trash/` print can address any more (trash purge cannot compute those names after the file is gone — they hash its mtime and size), only for files this module minted and only past a 24 h grace.
- **Schema v31**: `idx_gen_dir_recency` expression index on `(output_dir, COALESCE(file_mtime_ms, created_at_ms) DESC)` — the gallery listing's exact filter and order — so `GET /api/gallery` walks rows in listing order instead of a temp B-tree sort of the whole directory on every poll.

## Guards

- `thumbnails.rs`: variant query defaults/rejections, default variant keeps the historical path and ETag, sweep removes only orphaned versioned tiles (keeps live variants, fresh tiles, TUI/waveform layouts).
- `migrations.rs`: `gallery_listing_uses_the_recency_index_instead_of_a_temp_sort` (`EXPLAIN QUERY PLAN` must name the index and no `TEMP B-TREE`).
- `routes.rs`: `thumbnail_warmup_observes_one_entry_atomically_and_yields_to_writer` unchanged and green.

## Docs

`website/api/index.md` (query params), `website/guide/configuration.md` (warmup row), `CLAUDE.md` (schema v31), `changelog.d/gallery-perf-server-thumbnails.md`.

## Test plan

- [x] `cargo fmt --all -- --check` (+ desktop crate), `cargo test -p mold-ai-db migrations`, `cargo test -p mold-ai-server thumbnail`, clippy `-D warnings` on server + db
- [x] Full `cargo test -p mold-ai-server` locally: 1 955 pass; the 13 that vary run to run (durable_queue_feeder timing tests, an H3 listing test) fail identically on unmodified `main` on this machine — environmental, see CI


jamesbrink:
Codex peer review (`codex review --base main`) findings, addressed in the follow-up commit:

1. **P2 — small prints demoting a whole origin.** The reply now carries `x-mold-thumbnail-rendition: <size>-<ext>` naming the rendition the server understood. The desktop's `should_downgrade_tier` demotes an origin to 256 px only when a 512 request comes back small AND undeclared (an older server); a current server answering a genuinely small print keeps retina for the host (test: `header_probe_reads_png_and_jpeg_dimensions`).
2. **P2 — malformed query defaulted.** `Query<ThumbnailQuery>` is no longer `Option`-wrapped, so `?size=abc` is axum's 400 while an absent query still parses (both fields optional).
3. **P3 — sweep racing a restore.** `sweep_orphans` runs under the gallery publication read gate, so a print cannot move between `.trash/` and the live dir between the two walks.

PR #1491, opened 2026-08-31
fix(ci): teach the desktop packaging guard about the cudnn feature

jamesbrink:
`main`'s CI has been failing for **eight consecutive commits**. The `FlashAttention feature typecheck` job runs

```
cargo check -p mold-ai-server --features h3-private-uat --all-targets
```

which fails with:

```
error[E0004]: non-exhaustive patterns: `&scheduler::PreparationEvent::Progress { .. }` not covered
```

The queue-progress work added `PreparationEvent::Progress`, but `ref2va_preparation_binds_references_from_its_own_hydration` still matched only `Ready` and `Failed`.

## Why nothing else caught it

The test is gated:

```rust
#[cfg(all(unix, any(feature = "h3", feature = "h3-private-uat")))]
```

so the default `rust` job compiles it out, and a full local `scripts/ci-local.sh` run does too. Only this one job builds that feature combination — a green PR genuinely could not have told anyone.

## The fix

The test also assumed the first preparation event was the terminal one, which stopped being true when `Progress` was introduced. It now skips any number of progress ticks (asserting each carries the same work id) and asserts on the outcome. Where `Progress` never fires the behaviour is identical, so this cannot change a case that passed before.

## Verification

| check | result |
|---|---|
| `cargo check -p mold-ai-server --features h3-private-uat --all-targets` (the exact CI command) | exit 0 |
| `cargo test -p mold-ai-server --features h3-cuda,nvml,mp4 ref2va_preparation_binds_references` | **1 passed** |

The test cannot run under `h3-private-uat` alone — the private ingress boundary returns 401 there, independently of this change — so it was executed under the reviewed H3 feature set instead.

Found while watching main's CI after #1485. Pre-existing and unrelated to that change.

🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_01AjPWQj3ZabxXoMRhptD4c9

### DIFF
--- a/scripts/tests/desktop-nightly-race-guards.sh
+++ b/scripts/tests/desktop-nightly-race-guards.sh
@@ -30,7 +30,7 @@ grep -Fq 'run: cargo test --manifest-path src-tauri/Cargo.toml' <<< "$rust_job" 
   || fail "the native desktop gate no longer runs the test suite"
 
 linux_job="$(sed -n '/^  desktop-linux:/,/^  desktop-nightly:/p' "$workflow")"
-grep -Fq 'bunx tauri build --features h3-cuda,pulid --bundles appimage --ci -v' <<< "$linux_job" \
+grep -Fq 'bunx tauri build --features h3-cuda,cudnn,pulid --bundles appimage --ci -v' <<< "$linux_job" \
   || fail "main pushes have no Linux packaging proof"
 grep -Fq "if: github.event_name != 'pull_request'" <<< "$linux_job" \
   || fail "Linux packaging is not reserved for main pushes"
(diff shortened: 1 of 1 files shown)


jamesbrink:
Added a second commit: `a44fa0bc`.

Main's `Desktop` workflow is also red, and **that one is my own regression from #1485**, not pre-existing:

- The packaging guard asserts the AppImage proof by grepping the exact tauri invocation, and #1485 added `cudnn` to those features, so the literal stopped matching.
- The same guard extracts the Nix desktop recipe with a sed range anchored on `desktopFeaturesFor = computeCap:`. nixfmt split that across two lines once the recipe grew, so the range captured nothing and the guard failed on its *"no longer builds face identity"* assertion instead of on the thing it guards. The range now starts at the binding name alone and survives reformatting.

Neither was reachable before merge: `scripts/tests/desktop-nightly-race-guards.sh` is run only by the Desktop workflow, gated to main pushes, and never by `scripts/ci-local.sh` — so #1485 was green on its PR *and* on a full local gate and still turned main red.

`bash scripts/tests/desktop-nightly-race-guards.sh` → `PASS`.

So this PR now unbreaks both of main's red workflows: `CI` (pre-existing, 8 commits) and `Desktop` (mine, from #1485).

jamesbrink:
Rebased and **scoped down to one commit**.

#1490 landed the Ref2VA `Progress` fix concurrently (`dc7aed48`) — an equivalent fix, with the timeout wrapping the whole loop rather than each receive, which is the better shape since it budgets the total wait. My redundant commit is dropped.

What remains is only the part that is uniquely mine: the desktop packaging guard, which **my own #1485 broke**.

- The guard asserts the AppImage proof by grepping the exact tauri invocation; #1485 added `cudnn` to those features, so the literal stopped matching.
- The same guard extracts the Nix desktop recipe with a sed range anchored on `desktopFeaturesFor = computeCap:`. nixfmt split that across two lines once the recipe grew, so the range captured nothing and the guard failed on its *"no longer builds face identity"* assertion rather than on the thing it guards. It now anchors on the binding name alone and survives reformatting.

`bash scripts/tests/desktop-nightly-race-guards.sh` → `PASS`, rebased on top of #1490.

Note for the record: `scripts/ci-local.sh` never runs this guard — only the Desktop workflow does, gated to main pushes — so #1485 was green on its PR *and* on a full local gate and still turned main red.

PR #1522, opened 2026-09-01
fix(cli): report on the selected host and stop dumping SQL

jamesbrink:
Two CLI diagnostics that answered about the wrong thing.

## `mold server status` ignored the host you selected

`run_status` read only this machine's PID file, so `MOLD_HOST=plato mold server status` answered **"No server running"** about a host it never contacted — even though `crates/mold-cli/src/skill/references/cli.md` already documents that exact invocation, and the CLI reference already states that `gpu`, `ps` and `unload` never answer with this machine's state when a remote host is selected. `server status` was the outlier.

`--host` (env `MOLD_HOST`) now routes through `status_target()`:

- an explicitly named non-local server → HTTP status read, reporting what the server itself knows (version, uptime, models, devices). No PID, port or log path: those describe a local process.
- a loopback host on the managed daemon's own port → unchanged local reading, including the unmanaged-process scan.
- a loopback host on a *different* port → probed over HTTP, since it isn't the managed daemon.
- unreachable host → `No server responding at <url>`, exit 1 (not "No server running").

Verified against a real remote (`plato`, 4× L40S):

```
$ mold server status --host plato
Server running at http://plato:7680
  Version: 0.26.0
  Uptime:  2528s
  Models:  minimax-h3-fl2va:comfy-pruned-int8, qwen-image-2512:q8, flux2-dev:q8
  GPU 0: NVIDIA L40S [cuda:9ffc81c5…] Enabled/Healthy, VRAM 11756/46068MB, util 100%
  …
  Status:  busy (generating)
```

## A failed migration dumped its whole SQL statement

rusqlite's `SqlInputError` displays as `{msg} in {sql} at offset {n}`, where `{sql}` is the entire statement being prepared. For a migration that is a whole `CREATE TABLE` block, so a DB stuck on a version conflict printed ~25 lines of DDL into the middle of an ordinary `mold run`:

```
WARN mold_db: metadata DB unavailable: applying migrations … table video_upscale_jobs already exists in
CREATE TABLE video_upscale_jobs (
    id                TEXT PRIMARY KEY,
    …25 lines…
 at offset 14: Error code 1: SQL error or missing database
```

Now:

```
WARN mold_db: metadata DB unavailable: applying migrations to metadata DB at …/mold.db: migration v34 failed (DB at v33): table video_upscale_jobs already exists
```

The statement text lives in `migrations.rs`; what the user needs is which migration, which version the DB is stuck at, and what SQLite objected to. Rust migrations get the same `migration v{n} failed (DB at v{m})` context.

(The stuck DB that surfaced this was a local artifact of running a pre-renumber branch build — `video_upscale_jobs` stamped as v33 before `gallery_media_*` took that number — not a state main can produce. The output was wrong regardless.)

## Tests

- `status_target` mapping: no host, remote host, loopback same port, loopback other port.
- `migration_failed` renders one line, keeps SQLite's complaint, contains no `CREATE TABLE`.
- `apply_pending` on a DB carrying a conflicting table fails with the concise diagnosis.

`cargo test -p mold-ai-db` (325 passed) and `--bin mold` (734 passed). Two pre-existing failures in `commands::rm` and `commands::info` fail identically on unmodified `main` here — macOS `/private/var` canonicalization and local `MOLD_HOME` state.

Docs: `website/guide/cli-reference.md` gains a `mold server status` section and the standalone-commands table row is corrected; the agent-skill CLI reference gains the `--host` form.

https://claude.ai/code/session_01Y1Yg5y9vjrvmqjApp4hRD4


jamesbrink:
Codex peer review (one round) — findings addressed in 2a4eb1f81.

**P1, remote status ignored `MOLD_API_KEY` — fixed.** `MoldClient::new` never reads the key, so an authenticated server answered `/api/status` with 401 and the command called it unreachable. The status client is now built through `control::client_for_host`, the CLI's one authenticated construction path — applied to the local managed daemon too, which had the same hole. Regression test uses wiremock (`x-api-key` required for 200, 401 otherwise) and fails against the old anonymous client; `report_remote_status` now returns whether the server answered instead of exiting from inside the probe, which is what makes it testable.

**P2, explicit loopback ports were swallowed — fixed.** `managed_port.is_none_or(...)` treated every loopback port as the managed daemon when no PID file existed, so `--host localhost:9999` ran the local process scan instead of probing the port the user named. Only the managed daemon's own address keeps the PID reading now; any other loopback port is probed over HTTP, with the unmanaged-process scan demoted to a fallback for when nothing answers there. Loopback detection reuses `control::is_loopback_host` (which also handles `127.0.0.2`, unlike my hand-rolled match).

**P3, one changelog bullet — not applied.** `changelog.d/README.md` says "Several bullets in one fragment are fine when one PR ships several notes," and this PR ships two independent user-visible changes. `CLAUDE.md` did say one bullet per PR, which is where the reading came from — the two docs disagreed, so this PR reconciles `CLAUDE.md` with the README rather than splitting the note.

PR #1524, opened 2026-09-01
fix(gallery): open retained source media on a keyless host

jamesbrink:
## Problem

Generate with the CLI against a remote host that runs without `MOLD_API_KEY` (plato), then **Reuse settings** on that print in the desktop app:

> Connect this machine with an API key to restore its private source media.

Nothing needs an API key, and the print was a plain text-to-image render with no source media to restore.

Reproduced directly:

```
$ curl -s http://plato:7680/api/gallery/source-media/this-file-does-not-exist.png
{"availability":"unavailable_auth"}   HTTP 200
```

The server answers `unavailable_auth` for a filename it never looks at.

## Root cause

`crates/mold-server/src/gallery_source_media.rs`

```rust
fn private_auth_enabled(auth: Option<&Extension<AuthState>>) -> bool {
    auth.is_some_and(|Extension(state)| state.is_some())
}
```

`AuthState` is `None` whenever `MOLD_API_KEY` is unset, so this asked **"does the server enforce keys?"**, not **"is this caller authorized?"** — permanently `false` on a default host. All three retained-media routes refused, so the feature shipped in #1498 was dead on every keyless server. This contradicts the auth model everywhere else: with no key configured `require_api_key` passes every request, and `DELETE /api/gallery/image/:filename` and device lifecycle are already open (`device_patch_is_open_when_server_auth_is_disabled`).

## Fix

**Server** — `private_media_authorized(auth, authenticated)`: on a keyed host the caller must carry the `ApiKeyAuthenticated` marker; on a keyless host every caller is authorized, like every other route. Reuse sessions on a keyless host bind one stable anonymous subject on both create and hydrate. A clean resolve with no downloadable members is `unavailable_legacy`, not `missing_or_corrupt` (H3 retains only `source_image_name`, which `downloadable_role` filters by design).

**Shared contract** (`studio/api/gallerySourceMedia.ts`) — `printRetainedSourceMediaCandidate(metadata)`: the server cannot tell a pre-feature print from one that never had source media (both resolve with no pins → `unavailable_legacy` → "Reattach it before developing"), so every client gates the probe on the print's own recorded conditioning bytes and never asks about a text-to-image print. The probe maps a middleware `401` to `unavailable_auth`, so a keyed host reached without a key gets the API-key disclosure instead of a swallowed error.

**Desktop** — Library gates the probe; the Lightbox's primary **Reuse these settings** button now emits to the owner's `reuseSettings` instead of calling `composer.set` (which *invalidates* retained authority — only the right-click item was attaching it).

**Mobile** — stops synthesising `unavailable_auth` from "this phone holds no key"; asks the host, gated the same way.

**Web** — Library gates the probe the same way.

## Tests

- Server: 5 new unit tests on the gate, the anonymous subject, and the availability mapping (there was no coverage of the gate at all).
- Studio: 5 new tests on the predicate and the 401 mapping.
- Desktop: 4 new `LibraryView` tests (t2i never probes / img2img attaches inventory / auth disclosure / Lightbox primary button == context item); `Lightbox` test updated to assert the emit.
- Mobile: 2 new `MobileApp` tests mirroring the reported scenario.
- Web: 2 new `LibraryPage` tests.

`bun run check:frontend` green (8,955 tests). `cargo clippy --workspace --all-targets -D warnings` and `cargo fmt` clean; `mold-ai-server` suite green.

## Docs

`CLAUDE.md`, `website/api/index.md`, `desktop/docs/server-api.md`, `apps/mobile/README.md`, `changelog.d/retained-source-media-open-host.md`.

## Not in scope

Web's Create-drawer recent-prints lightbox (`recreateFromGallery`) restores metadata but never attaches retained source media — a separate parity gap, not the reported bug.

https://claude.ai/code/session_019b5TChbaNUnfwYchdqcgmK


jamesbrink:
**Codex review (1 finding, P1) — addressed in 5e176a7.**

> Gating the probe on metadata markers misclassifies prints conditioned with inline `source_video`, `audio_file`, `extend_video`, or a standalone `control_image` as having no retained media, so Reuse settings silently loses those inputs.

Confirmed from `OutputMetadata::from_request`: inline video/audio/mask bytes leave no marker (only the `_path` forms are recorded, and browser clients always send bytes). Reworked so every surface **always probes** — the server is the only authority on what it retained — and the metadata predicate (`retainedSourceMediaDisclosable`, now also reading `extend_overlap_frames`) decides only whether an *unavailable* answer is worth a toast. That keeps the text-to-image print quiet without losing any restoration. Tests updated; added a desktop case for a V2V print with no marker.

Also fixed in the same push: the two red checks (`Frontend` prettier scope for `desktop/docs/`, and `docs` — table realignment plus the pre-existing `MOLD_BUNDLED_FFMPEG`/`FFPROBE` verify gap from #1506, which would have failed the docs job for any PR that runs it).

PR #1558, opened 2026-09-02
feat(h3): print-ready mesh export geometry and a turntable framed once per sweep

jamesbrink:
## Why

A Discord user loaded an STL exported by mold into Bambu Studio and got "Object too small … may be in meters or inches". Hunyuan3D meshes are stored in normalized model space (unit cube, Y-up) and the export transcoded them verbatim, so slicers saw a 2 mm blob and Blender saw the model on its side. Separately, the turntable GIF/APNG/WebP "zoomed in and out": the rasterizer refit its scale to every frame's own silhouette, so a rotating mesh breathed by up to ~41 % and popped at the x/y crossover.

## What

**Geometry export options** (`size_mm`, `up_axis`, `origin`) on OBJ/STL/PLY exports, from one server-advertised per-format defaults table (`mold_core::validation::mesh_export_geometry_defaults`):

| Format | Size | Up | Origin |
| --- | --- | --- | --- |
| STL, PLY | 100 mm longest axis | Z-up | floor (bed-ready) |
| OBJ | as stored (model units; DCC tools treat 1 unit = 1 m, Blender's OBJ importer converts Y-up itself) | Y-up | floor |

- Advertised as `capabilities.mesh.export_geometry`; its presence is the only client gate (an older server silently ignores the keys, so clients never send them without the block and `mold library export` refuses the flags against such a host).
- Refused (422 / CLI error / MCP error) on `glb` and on turntable and video formats, never ignored.
- Every surface: `POST /api/gallery/export/:filename`, `mold library export --size-mm/--up-axis/--origin`, the `export_mesh` MCP tool, the TUI's local writer (byte-identical to a served default export), and a shared options sheet on web and desktop with inline fields in the phone's swipe-up sheet.
- `Mesh::apply_export_geometry` recentres on the mesh's own bounding box (the extraction centres on the grid), scales, rotates positions and normals (det +1, never a mirror), then floors.

**Turntable framed once per sweep**: `raster::FrameFit { Auto, Extent }` on `Camera`; `turntable_frame_cameras` computes one extent over every camera in the sweep (mirroring ComfyUI's splat turntable, `nodes_gaussian_splat.py:996-1006`). The poster keeps `Auto`, so poster bytes are unchanged; frame 0 is now the poster's camera at the sweep's scale rather than its exact pixels (docs updated).

Docs: website API/guide/CLI/model pages, skill corpus, desktop docs, CLAUDE.md bullet, `changelog.d/mesh-export-geometry.md`.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check`, `generate_generation_profiles --check`, `ci-routing-contract.sh`, skill-corpus bash-fence test, frontend `check:frontend` / `vue-tsc -b` / `check:architecture` / `check:dead-code` / prettier: all clean.
- New tests: constant `projection_scale` across a sweep (with a non-vacuity guard that `Auto` varies > 25 %), no frame clips the edge, loop seamless, swept extent within the closed-form cylinder bound; mesh transform (scale, rotation + normals + STL facet normals, floor, recentre, degenerate, no-op byte identity); server 422s and capability block; CLI/MCP parse and refusals; TUI local/served parity; old-server regression on web, desktop and mobile.
- Live UAT with a real print pulled from plato, served by a build of this branch:
  - default STL: 100.0 × 74.3 × 88.6 mm, Z-up, min z = 0.000
  - `--size-mm 120 --up-axis y --origin center`: 120.0 × 106.3 × 89.1 mm centred
  - default OBJ: unscaled 1.96 × 1.74 × 1.46, Y-up, min y = 0
  - same print on plato (main): 1.96 units across, reproducing the bug
  - `--size-mm` against plato: refused with the host-does-not-advertise message; plain export unchanged
  - 36-frame turntable from both servers: the branch holds one size through the turn; main visibly breathes.

Follow-up (not in this PR): surface `frames` in the video export sheet; desktop Lightbox shows meshes through `AuthedMedia` so its size line reads "longest side" instead of three extents.

https://claude.ai/code/session_01LnEtwgN2NkwLc4ui8U5pC7


jamesbrink:
Peer review (Claude sub-agent, opus) found one must-fix and two wording issues, all addressed in 07825c99c:

1. **MCP `export_mesh` had no capability gate** — it was the only client that could send `size_mm`/`up_axis`/`origin` to a host that never advertised the block. An agent asking for a 120 mm STL from a pre-feature server would get an unscaled Y-up file back with a success message and nothing to catch. It now probes capabilities and refuses by name, only when a geometry argument was actually set. The refusal sentence moved into `mold_core::validation::mesh_geometry_unadvertised_refusal` so the CLI and MCP wordings cannot drift, and a new test asserts no export request goes out (wiremock `.expect(0)`).
2. **CLI help, tool description and TUI hint** stated the print-ready defaults as unconditional facts; the capability gate only covers explicit flags, so a default export against an old host still returns an unscaled mesh. The strings now qualify that with "on a host that advertises them".
3. **"OBJ as stored"** was inaccurate in the TUI hint: OBJ is unscaled but still recentred and floored, so it now reads "OBJ keeps its model units".

Also formats the two markdown files that failed the `docs` and `Frontend` jobs. The desktop and website prettier scopes are wider than the repo root's, which is why the root `fmt:check` passed locally.

Nits not taken: `FrameFit::Extent` uses `min(half_w, half_h)` while `Auto` uses per-axis extents, which cannot differ today because turntable frames are always square.

https://claude.ai/code/session_01LnEtwgN2NkwLc4ui8U5pC7

PR #1580, opened 2026-09-04
test(desktop): bound the engine boot shutdown by the server's own budget and print the elapsed time

jamesbrink:
## Summary

`desktop/src-tauri/tests/engine_boot.rs` has been failing intermittently on the macOS **Native Rust** job with `engine thread did not exit`, always at 15.6–16 s (the test's own bound): main runs 33616295687 (Sept 2), 33820692419, 33823114932, PR run 33821746457, and feat/six-themes 33799762208. Each failure on main skips **Publish nightly update**, which is what kept tonight's desktop nightly (carrying #1577) from going out.

The bound was stricter than the contract it exercises: `mold_server::run_server` waits up to `DEFAULT_SHUTDOWN_ABORT_SECS` (45 s, `crates/mold-server/src/lib.rs:1482`) for GPU owners before it stops waiting (`lib.rs:1590`), so an engine that exits inside that budget is behaving. Locally the whole shutdown takes 55 ms; on a loaded runner one of the awaited shutdown steps evidently takes longer than 15 s.

## Change

- Join bound = `DEFAULT_SHUTDOWN_ABORT_SECS + 15` (60 s).
- Print `engine shutdown took … (bound …)` on every run, and include the elapsed time in the failure message, so a run drifting toward the budget is visible in the CI log before it fails and the next failure names its cost.

A genuine hang still fails (at 60 s instead of 15 s). Test-only; `skip-changelog`. Passes locally.

### DIFF
--- a/desktop/src-tauri/Cargo.lock
+++ b/desktop/src-tauri/Cargo.lock
@@ -4198,6 +4198,7 @@ dependencies = [
  "time",
  "tokio",
  "tracing",
+ "tracing-subscriber",
  "uuid",
 ]
 
--- a/desktop/src-tauri/Cargo.toml
+++ b/desktop/src-tauri/Cargo.toml
@@ -149,6 +149,9 @@ codegen-units = 16
 strip = true
 
 [dev-dependencies]
+# Test-only: the engine-boot test prints the server's own shutdown log so a
+# stalled phase names itself in CI instead of only reporting elapsed time.
+tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 # Test-only: builds the two-triangle GLB the offline mesh-poster tests render.
 # `default = []` and the crate is already in this root's graph through
 # mold-server, so this edge adds no features and no extra compilation.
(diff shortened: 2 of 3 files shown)


jamesbrink:
Update: the widened bound did **not** pass. The macOS runner failed again at **60.2 s** against the 60 s bound, so this is a real stall in embedded shutdown, not runner jitter, and the original 15 s bound was reporting a genuine defect.

Repointed this PR at diagnosis: the test now prints the server's own shutdown log to stderr and writes the elapsed time through the stderr handle (libtest captures the print macros and shows them only for a failing test, so a slow-but-passing run said nothing). The next CI failure will name the phase that stalls.

Locally the whole sequence takes 55 ms and ends at the durable-queue feeder. Candidates for the CI stall, in order: the mDNS browser guard's unbounded `thread.join()` after `daemon.shutdown()` (`crates/mold-server/src/mdns.rs:70-77` — the browse receiver's `recv()` blocks forever if the daemon never delivers `SearchStopped`, which is plausible on a runner without usable multicast), then Axum's graceful drain waiting on the test client's idle keep-alive connections.

Worth noting for whoever picks this up: embedded shutdown has **no** enforced deadline (the hard-exit deadline is deliberately off for an embedded server), while `stop_local_engine` gives it 10 s before telling the user gallery authority is stuck with the server. So whatever this is, it is a user-facing hang on macOS, not only a CI annoyance.

PR #1581, opened 2026-09-04
fix(mdns): bound the peer browser's shutdown by mold, not by the mDNS daemon

jamesbrink:
## The bug

`MdnsBrowserGuard::shutdown` (`crates/mold-server/src/mdns.rs`) joins the browse worker thread with no bound. That thread sat in a blocking `recv()` on the browse channel, whose sender belongs to the mDNS daemon — so the only thing that could wake it was the daemon closing the channel. On a host where multicast never works, it may never close, and the join never returns. `run_server` never returns either.

That is user-facing, not just CI: `stop_local_engine` gives the embedded engine 10 s before telling the user *"The embedded engine did not stop within 10s; gallery authority remains with the server."*, and the hard-exit deadline is deliberately off for an embedded server (it would take the whole app down).

## How it surfaced

`engine_boots_authenticates_and_shuts_down` has been failing intermittently on the macOS **Native Rust** job, always at exactly its own bound (main runs 33616295687, 33820692419, 33823114932; PR run 33821746457). Raising the bound to 60 s produced a **60.2 s** failure, which ruled out runner jitter: something genuinely never finishes. Each failure on main skips **Publish nightly update**, which is why the desktop nightly stopped arriving.

The browser join is the only unbounded wait in that shutdown sequence, and the last log line before the stall is consistent with it.

## Fix

The loop polls `recv_timeout(BROWSE_STOP_POLL)` (250 ms) against a stop flag the guard sets *before* asking the daemon to stop. A responsive daemon still ends the loop instantly by closing the channel; an unresponsive one now costs one poll interval instead of the process.

## Tests

- `the_browse_loop_stops_on_the_flag_even_when_the_channel_stays_open` — live sender, no events, flag set. **Verified failing against the old blocking `recv()`** and passing with the fix.
- `the_browse_loop_stops_when_the_daemon_closes_the_channel` — the fast path still ends immediately.

`cargo clippy -p mold-ai-server --features mdns --all-targets -D warnings` and `cargo fmt --all --check` clean. Related: #1580 (test diagnostics that located this).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -637,6 +637,16 @@ jobs:
           cargo test -p mold-ai-core --features pulid --lib identity
           cargo test -p mold-ai-inference --features pulid --lib identity
           cargo test -p mold-ai-server --features pulid --lib identity
+      # `mdns` is not a default feature and nothing in the workspace turns it
+      # on, so `cargo test/clippy --workspace` never COMPILES mdns.rs — the
+      # module's tests, including the guard against reintroducing the blocking
+      # `recv()` that hung server shutdown (#1581), were dark. The desktop app
+      # enables the feature but is a separate cargo root.
+      - name: Clippy and test the mDNS surface
+        if: env.RUN_RUST_SUITE == 'true'
+        run: |
+          cargo clippy -p mold-ai-server --features mdns --all-targets -- -D warnings
+          cargo test -p mold-ai-server --features mdns --lib mdns
       - name: Record protected gate success
         run: echo "Required Rust and release routes completed successfully."
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1615,6 +1615,9 @@ name = "fastrand"
 version = "2.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
+dependencies = [
+ "getrandom 0.3.4",
+]
 
 [[package]]
 name = "fax"
@@ -1723,6 +1726,7 @@ version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5e139bc46ca777eb5efaf62df0ab8cc5fd400866427e56c68b22e414e53bd3be"
 dependencies = [
+ "fastrand",
  "futures-core",
  "futures-sink",
  "spin",
@@ -3452,6 +3456,7 @@ dependencies = [
  "cudarc 0.19.8",
  "dirs 5.0.1",
  "filetime",
+ "flume",
  "fs2",
  "futures",
  "futures-core",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1615,9 +1615,6 @@ name = "fastrand"
 version = "2.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
-dependencies = [
- "getrandom 0.3.4",
-]
 
 [[package]]
 name = "fax"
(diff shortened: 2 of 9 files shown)


jamesbrink:
Confirmed on main while this was in review: run 33826204557's macOS Native Rust job hit **engine shutdown took 60.0s** again, so the stall is live and ongoing, not a one-off.

Review follow-ups pushed:
- **The new tests would never have run.** Nothing in the workspace enables mold-server's non-default `mdns` feature, so `cargo test/clippy --workspace` never compiled `mdns.rs` at all — the regression guard was dark, and the clean clippy bill came from a local run. Added a feature-scoped clippy + test step to `.github/workflows/ci.yml` and `scripts/ci-local.sh`.
- **`start_browser`'s doc comment** had been left attached to the new private helper, leaving a `pub fn` in a `pub mod` undocumented. Moved back.
- **`flume` dev-dep now mirrors mdns-sd's `default-features = false`**, so the test channel behaves like the daemon's (default `eventual-fairness` changes receiver wake selection) and `Cargo.lock` is byte-identical to main again.
- Dropped the redundant `own_instance_id.clone()`.

Left the 250 ms poll as is: halving the wake-ups would double worst-case shutdown latency, and 4 wake-ups/sec is well under the cost of the hang it replaces.

jamesbrink:
**Correction: this does not fix the hang.** With the mDNS bounded shutdown in place, the macOS job failed again at 60.1 s (PR run 33827438550), and I confirmed mdns-sd's `shutdown()`/`stop_browse()` are non-blocking `try_send` calls, so the daemon was never the stall. The bounded join and its tests stay — an unbounded join on a shutdown path is a real defect, and the guard plus the new CI coverage keep it fixed — but the PR no longer claims to cure the hang.

What the log actually shows: the server reaches the durable feeder and then goes silent, in both passing and failing runs, so nothing distinguished a dozen consecutive awaits. Added two layers of diagnosis:

1. `run_server` now names each shutdown step at debug level (`awaiting shutdown step step=…`, then `shutdown sequence complete`).
2. The desktop engine thread names the boundary after that: `run_server returned; dropping the runtime` and `runtime dropped; thread exiting`.

That second boundary is the leading suspect. Dropping the Tokio runtime deliberately waits for every blocking task (`shutdown_timeout` was rejected because it detaches them, and thread exit is what proves the server no longer owns the gallery). One hung `spawn_blocking` — a `statvfs` that never returns is a known shape here, see the leaked-blocking-thread comment in `state.rs` — parks the engine thread forever with the entire shutdown sequence already logged as complete. That matches the symptom exactly: every bound we set, 15 s then 60 s, is hit precisely, which is an infinite wait being truncated rather than a slow step.

The next failure will say which side of that boundary it is on.

jamesbrink:
**Found it, and it was the test.**

The instrumented run (job 100885718059) is unambiguous. Every one of the thirteen shutdown steps logged at 02:14:19.222–19.225, and the test's own timeout fired at 02:14:19.155. The entire shutdown sequence ran within **2 ms of the test giving up and unwinding** — it had not started before that.

What unwinding does is drop the test's `reqwest::Client`, and with it the pooled keep-alive connection to the server it was waiting on. Axum's graceful shutdown drains live connections, so the test was itself the reason `run_server` could not return. That explains every observation: the failure always lands exactly on whatever bound is set (15.6 s at a 15 s bound, 60.1 s at 60 s), because it is an unbounded wait being truncated, and it is intermittent because it depends on whether that pooled connection is still alive when shutdown begins.

Fix is one line plus the reason: drop the response and the client before the join.

Everything else in this PR stays and is still worth having:
- the mDNS browser's unbounded join is a real defect on a shutdown path with a 10 s user-facing budget, now bounded and pinned by a test that fails against the old blocking `recv()`;
- `mdns.rs` was never compiled by any workspace CI gate, so that test would have been dark — the feature-scoped clippy + test step fixes that;
- the per-step shutdown trace and the runtime-drop boundary are what turned an opaque hang into a two-minute diagnosis, and they stay in for the next one.

Worth a follow-up issue, not this PR: if an idle client connection can hold the drain open, the desktop app's own `stop_local_engine` faces the same shape against its 10 s budget.

Issue #1597, opened 2026-09-05
Desktop redesign: server fields the client is waiting on


jamesbrink:
Split out of #1586 so the server work is tracked apart from the desktop's optional design changes. Every item here is a **field or endpoint the server does not have yet**; the client half is already designed (mock lines in `docs/design/mold-studio-desktop.dc.html`) and waits on it. None blocks anything shipped.

## Generation duration

- [x] **Per-print generation time in `OutputMetadata`.** `generation_time_ms` exists only on the live complete event and on queue jobs, so a print reloaded from the gallery has no duration. Unlocks the Recent tab's `flux-dev:q4 · 4.0s`, the canvas caption's `· 4.0s`, the Lightbox's "Took 4.0s", and a time estimate on the Quality rows.
- [x] **SPEED column in Styles** (mock L784, `~20s` / `~3s` per style): `ModelEntry` has no timing field. Derive from recent prints once the field above exists, or advertise a measured figure per style.

## Rental price

- [x] **Hourly rate before renting a GPU.** `RunPodGpu` carries no price and the Rust side does not fetch one, so the "billing begins now" confirm cannot state a number. Worth an API addition.

## Storage totals

- [x] **Picture bytes in a machine's Storage card** (mock L921 "pictures take 12.4 GB"): no gallery-bytes total on `/api/status` or `/api/capabilities`; `size_bytes` exists per row only.

## Other server fields

- [x] **Save every result** toggle in the New image inspector (mock L528–531): the generate request has no "don't persist this print" flag. Needs a server-side flag and admission rule before a control can ship.
- [x] **Friendly style names on licence rows**: the licence listing returns `name` + `summary` only; the row shows the id in mono.
- [x] ~~**Pausing the job that is already running**~~ — **closed as not possible.** An in-flight denoise cannot be suspended by any engine here, so there is nothing for an API to call. The active card's pause holds the queue after the current image and says so; recorded as the one remaining named absence in `docs/design/README.md`.
- [x] **`GET /api/gallery/image/:filename` has no `view=trash` switch**, so a client that wants the `.trash/` bytes of a print that also exists live cannot ask for them. Desktop hides every reuse door on a trashed print instead, so nothing reaches this today.
[...]

Issue #1586, opened 2026-09-04
Desktop redesign follow-ups: backend fields, shared-surface vocabulary, deferred features


jamesbrink:
Follow-ups left over from the desktop redesign parity work (#1583), each named in a batch report or the design README's divergences rather than silently dropped. None blocks anything shipped.

**Reorganised 2026-09-05:** everything that needs a server field or endpoint moved to #1597 (generation duration, rental price, storage totals, Save every result, licence names, pausing the running job, `view=trash`, the H3 bridge ladder). What stays here is client-side: bugs found in UAT, shared-surface vocabulary that waits on web and the phone, and design choices deferred on purpose.

## Bugs — fixed on `fix/desktop-ui-fixes`

- [x] **The palette's "Make a short clip" deep-linked `/create?output=sequence`**, which did nothing from inside New image (`consumeOutputQuery` runs only on mount) and landed in Scenes when the door opens onto Simple. The row and File ▸ New Clip now raise a `shortClip` intent the view answers with the toolbar's own door (`useOutputKindDoor`), so they open Short clip on the remembered way and style from anywhere.
- [x] **The shared validator said "Describe clip 1 before generating." on desktop**, where the word is scene. `sequenceValidation` takes the caller's wording (`SequenceLimits.wording`); desktop passes `scene` / `clip`, web and the phone keep `clip` / `sequence`. Seen twice in UAT.
- [x] **`ConnectMachineModal.test.ts` was order-fragile.** The cause was not the appPrefs side effects but the network: a connect's status poll reaches `fetch` through the shared `listQueue` / `listDevices`, and `hal9000` and `192.168.1.30` are real machines on the developer's LAN, so whether they answered decided the "generation target" case. `fetch` is stubbed in that file.

## Shared code that web and the phone also render

- [x] **Literal radii and font sizes in shared `ui/` and `studio/` primitives** (~250 declarations across ~45 files, ratcheted per file in `desktop/src/styles/tokens.legacy.test.ts` `UNCONVERTED`): convert to the theme scale when web and the phone take the redesign, since values below 11px have no token and moving them blind shifts three surfaces.
- [x] **Web and the phone still render `meta.tone`** ("dark"/"light") in their theme pickers; `ThemeMeta.toneLabel` carries the phrases when their redesign lands.
[...]

Issue #1583, opened 2026-09-04
Desktop redesign phase 2: parity with the Mold Studio mock


jamesbrink:
## Summary

Phase 1 of the Mold Studio desktop redesign rebuilt every desktop view on the new design system (six themes, the `--mold-*` token layer, the plain-English lexicon, the new shell) on `feat/ui-redesign-desktop`. A five-area design-parity audit then compared the branch against the reference implementation, `docs/design/mold-studio-desktop.dc.html`, and found **53 must-fix gaps, 48 should-fix, and 23 polish items** plus ~45 clutter items. This issue tracks closing them so the app matches the mock.

Design source of truth: `docs/design/README.md` (lexicon §2, shell anatomy, component vocabulary) and `docs/design/mold-studio-desktop.dc.html`. The audit reports are attached at the bottom.

## Decisions (made 2026-09-04)

All nine structural calls are **yes**:

1. **New image inspector** — delete the duplicate "Output · One shot | Sequence" card (the header's Still picture | Short clip | 3-D object control is the one output switch); add Draft / Good / Best Quality rows built from the recipe's steps range, driving Detail; lift Add-on looks out of Advanced; one Clip card (Length · Smoothness · Add sound + "you'll get a still preview first"); remove the style-preset strip from the desktop composer (it collides with the bound word "Style").
2. **Clip mode and timeline** — the composer stays mounted in clip mode (placeholder "Scene N — describe what happens next", chip "Make 1 clip"); a transport (play/pause + mono `0:06 / 0:20`), tick ruler, and playhead; a new desktop `SceneLane.vue` where a block's width equals its duration, the thumbnail sits behind, the seam chip floats on the join, and the selected scene carries the trim grip. The shared `ui/components/ClipRail.vue` stays for web. One word, "scene", everywhere.
3. **Starters / Recent tabs** — Starters as picture cards from the template store (the manager behind an edit affordance); Recent as recent prints with thumbnails and mono `model · time`, restoring full settings.
4. **Canvas caption** — Save · Make 4 variations · Make bigger · ⋯ (Start from this moves into ⋯); filename left, `1024² · 4.0s` right.
[...]

PR #1598, opened 2026-09-05
feat(desktop): Simple | Scenes strip, one vocabulary for the three kinds, per-section style memory

jamesbrink:
## Summary

Desktop UI fixes on top of the Mold Studio redesign (#1595), plus the three prioritised #1586 bugs and a CLAUDE.md repair.

- **Simple | Scenes has a row of its own.** `ClipModeStrip` renders how a clip gets made in a strip under the New image toolbar (the chip row's chrome), only while the kind is a clip, with one sentence saying what the chosen way does. Beside the kind control it pushed the whole right-hand cluster left whenever Short clip was chosen; `CreateHeader` now holds the same children in every kind, pinned by its test.
- **Styles filters by the three kinds Create offers, in the same words.** All · Still picture · Short clip · 3-D object from one label table (`OUTPUT_KIND_LABEL`), sorted by the same `outputKindForModel` partition the sections use; `?type=mesh` joins `image` / `video`, and Browse more from the 3-D section lands on it filtered.
- **Each section remembers its last-used style across restarts.** `studio/stores/lastUsedStyles.ts` (localStorage, like the draft) keeps the style each section was last used with and the section used last. Desktop's `GenerateView` is the one writer; every section door falls to the remembered style before the first installed one; the first visit of a launch lands on the style and section left, waiting for a ready machine still reporting before giving up on it, and stands down in Scenes where the sequence watcher owns the pick. Web and the phone adopt the same store with their redesigns (specs updated).
- **Make a short clip opens the Short clip door** (#1586). The palette row and File ▸ New Clip raise a `shortClip` intent the view answers with `useOutputKindDoor(...).setOutputKind("clip")` — the toolbar segment's own door, extracted from the header — instead of deep-linking Scenes and doing nothing from inside New image.
- **Scene validation speaks desktop's word** (#1586). `sequenceValidation` takes the caller's wording; desktop passes `scene` / `clip`, web and the phone keep the default.
- **`ConnectMachineModal.test` no longer reaches the real network** (#1586). The status poll after a connect goes through the shared `listQueue` / `listDevices` to `fetch`, and `hal9000` / `192.168.1.30` are real machines here; `fetch` is stubbed.
- **Desktop test setup clears `localStorage` before every test.** happy-dom keeps one per file and the new memory hydrates a fresh Pinia from it.
- **CLAUDE.md**: the Config section's body was orphaned under the gallery section; single-test commands added.

Docs: design README, desktop design canvas (strip moved under the toolbar), website desktop guide, `.claude/rules/{desktop,studio-web}.md`, `changelog.d/desktop-section-memory.md`.

Issues: #1583 reconciled (batch I ticked; native eyeball pass recorded), #1586 reorganised with the three bugs ticked, server-side items split to #1597.

## Test plan

- [x] `cd desktop && bunx vitest run` — 497 files, 6694 tests
- [x] `bunx vitest run --config studio/vitest.config.ts` — 1801
- [x] `bun run test:web` — 1872; `ui/` — 382
- [x] `cd desktop && bunx vue-tsc -b`; prettier on `desktop/src`, `studio`; `bun run check:architecture`; `bun run check:dead-code`
- [x] Codex review round (`codex review --base main`): two P2 findings, both real and fixed in `3b9a61f9` — the first-visit restore called the fleet settled while a remembered machine was still `connecting` at launch (`allReadyHostsFetched` counts only ready machines), and the `shortClip` intent raised from another workspace was consumed during setup before the inspector ref existed, dropping the Scenes swap. Both pinned by tests.
- [ ] Eyeball in `desktop-dev`: choose Short clip and confirm the kind control stays put; switch Simple ↔ Scenes on the strip; pick a clip style, switch to Still picture and back, quit, relaunch, confirm the clip style and section return; Styles ▸ kind filter reads Still picture · Short clip · 3-D object; ⌘K "Make a short clip" from inside New image.

https://claude.ai/code/session_011wBrhCYjqrWqnVA3D7Xb15


jamesbrink:
Addressed Greptile findings in `a0ee2748`:

| 1 | P1 | Doors could select a downloaded style no machine can run (`targetModels` keeps them for disclosure) | Fixed — `useOutputKindDoor` partitions only runnable rows (`isModelRuntimeUnavailable` filtered out), so neither the remembered style nor the section's first can be one; test "never opens a door onto a style no machine can run". |
| 2 | P1 | An offline machine kept `connecting` forever held the first-visit restore, so Create sat with no style | Fixed — "settled" is `hosts.initialized` plus every ready machine fetched, not "no machine connecting"; a failed boot probe keeps the row connecting by design. Tests cover both the reconnect wait and the offline fallback. |
| 3 | P2 | `shortClip` intent lost when starter cards hold the view and no inspector mounts | Fixed — the swap goes through the inspector when it has `setOutputMode`, otherwise switches the draft directly (nothing installed means nothing to swap; Scenes owns that empty state). Test asserts the starter-cards state and the draft in `sequence`. |
| 4 | P2 | Memory filed a style under `form.family`, which an edit session can leave stale | Fixed — the writer looks the name up in the installed inventory and files it under that entry's section; a name no machine has is not remembered. Test covers the stale-family and unknown-name cases. |

Full desktop suite 6700 green, vue-tsc, prettier, knip.

PR #1600, opened 2026-09-05
feat: record how long a print took and show it on desktop

jamesbrink:
## Summary

First of the #1597 series: the generation-duration field and everything it unlocks.

- **`OutputMetadata.generation_time_ms`** (additive). Set through one helper, `record_generation_time`, on every publication path: the GPU worker's completion, the single-worker CPU fallback, and the summed stages of a stitched chain print. Zero is the wire's "not measured" and leaves the field absent. `GenerationRecord::to_gallery_image` fills it from the gallery row's existing `generation_time_ms` column for a print made before the field, so legacy prints answer too.
- **Desktop reads it in one spelling** (`studio/lib/generationTime.ts`: `4.0s` under a minute, `1m 12s` above): the canvas caption (which used to print `0.0s` for a restored print), the Recent row (`flux-dev:q8 · 4.0s`), and a new **Took** fact in the Lightbox. A print that does not know shows nothing.
- **Styles gains the mock's Speed column**, derived rather than served: `studio/lib/styleSpeed.ts` takes the median of the newest ten timed prints per style (`~20s`), computed once per gallery change; an untimed style shows an empty cell so the pinned axis holds. The shelf is a size container and folds the column away below 60rem, where the fixed tracks would leave the name track with the residency star alone at the app's minimum window.
- Web's `OutputMetadata` type gains the field; the design README's named-absences list drops the duration and SPEED entries.

## Codex review

Two P2 findings, both real, fixed in the second commit: the CPU fallback worker built its metadata without the duration, and the new fixed tracks would have hidden style names at 1080px with the sidebar open.

## Test plan

- [x] `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check`
- [x] `cargo test -p mold-ai-core` (metadata additivity), `-p mold-ai-db` (row fallback), `-p mold-ai-server --lib queue::` / `gpu_worker`
- [x] Desktop 6714 tests, studio 1811, web 1872; vue-tsc, prettier, knip
- [ ] `chain_job_runner::tests::amend_later_stage_preserves_completed_source_bearing_prefix` fails locally on `main` too ("queue-media master key is missing while stored media exists") — environment, not this change; CI is the authority.

Refs #1597

https://claude.ai/code/session_011wBrhCYjqrWqnVA3D7Xb15


jamesbrink:
Addressed Greptile finding in `8490db9a`:

| 1 | P2 | The folded axis was declared on `.model-table`, which its own container query cannot restyle, so the Speed cells hid while their track stayed | Fixed — the five-track axis is set on the header and on every row inside the query; the shelf test now asserts that shape and refuses a `.model-table` declaration inside the query. |

PR #1608, opened 2026-09-06
fix(inference): recover Z-Image Metal VAE decode

### LINKED ISSUE #1040: candle Metal: silent VAE decode corruption past latent 120 — root-cause the allocator, remove the Z-Image tiling cap
## Summary

candle's Metal backend silently corrupts the Z-Image (FLUX-architecture) VAE decode past latent 120 per axis — wrong tensor, no error, even across a device synchronization. mold works around it by proactively tiling the decode on Metal (#946, now reduced to a single-pass 2×2 grid). This issue tracks the root-cause fix in `utensils/candle` so the workaround — `MAX_WHOLE_VAE_DECODE_LATENT_AXIS` in `crates/mold-inference/src/zimage/pipeline.rs` — can be deleted and Metal decodes 1024²+ whole (~6 s instead of ~9 s tiled, and no tiling seam risk at all).

## Symptom (measured, M4 Max 48 GB)

- Whole Z-Image VAE decode on Metal vs CPU reference: exact (rel ~2e-5) at latent 96/104/112/120; **rel error 1.61, partial data or all zeros at latent 128** (the 1024×1024 render). Bisection record in #946.
- Every constituent op measured exact in isolation at those shapes — but note the isolated conv was the 1.2e9-element one; the 2.4e9-element conv was never isolated.
- Failure is silent: command buffers report success.

## Root-cause analysis (2026-08-14 investigation)

Confirmed by code reading in `utensils/candle`:

1. **`buf_size` rounds every allocation to the next power of two** — `candle-core/src/metal_backend/device.rs:352`. The decoder's three largest im2col buffers (e.g. 256ch·3×3 conv at 1024²: `1024² × 2304 × 4 B = 9.00 GiB`) all cross their power-of-two bucket (8 → 16 GiB) at **exactly latent 121**, matching the 120-good/128-bad bisection.
2. **The pool never shrinks mid-graph.** `drop_unused_buffers` runs only from `wait_until_completed` / `flush_and_wait_current` (device.rs:131/175/185), which fire only on CPU readback or explicit synchronize — a VAE decode has neither, so transient buckets accumulate. Estimated live pool at latent 128 F32: ~35 GiB vs `recommendedMaxWorkingSetSize` = 37.44 GiB on a 48 GB M4 Max.
3. **No limit is ever checked.** `maxBufferLength` and `recommendedMaxWorkingSetSize` are bound (`candle-metal-kernels/src/metal/device.rs:134,138`) with zero call sites. `ResidencySet::insert_batch` discards `commit()`'s result (`residency_set.rs:44`); `requestResidency` runs once at construction. Residency that cannot be satisfied yields undefined GPU reads with **no** command-buffer error — matching the symptom exactly.
4. Empirical: a weights-free repro of the decoder conv ladder (`tmp/vae-metal-repro/`, kept out of the workspace) is **bit-exact at latent 128** in a clean process — so a pure 2³¹ index overflow in im2col/gemm is refuted (also refuted by reading: `conv.metal` uses `uint` tid + `size_t` interior indices; MLX gemm casts to `size_t` before stride math). The full-decoder repro shows the decode's transient footprint exceeds ~38 GB even at latent 120 — the corruption is a *pool-pressure* failure, which is why ops pass in isolation (fresh process, small pool) and the composition fails.

Latent second bug found while reading: `uint tid` in the im2col kernel genuinely wraps past 2³² elements — latent ≥ 171 (1368 px) — currently masked because the 32 GiB bucket exceeds `maxBufferLength` first.

## Discriminating experiment (fold into the fix's test suite)

`MOLD_VAE_DTYPE=bf16` whole decode at latent 128 on Metal: element count unchanged, bytes halved, pool ~18 GiB. Allocator hypothesis predicts it succeeds; an element-count overflow predicts it still fails. Run with ≥ 40 GB free (the box OOMs otherwise — mind other processes).

## Proposed fix (in `utensils/candle`)

1. Cap power-of-two bucketing: `size.next_power_of_two()` up to 64 MiB, then round up to 64 MiB alignment. Takes the four big buckets at latent 128 from 32.2 GiB → ~18.3 GiB — below the latent-120 total that already renders correctly. `find_available_buffer` already does smallest-fit-≥-size, so reuse degrades only in precision.
2. Reject `size > device.maxBufferLength` with a real `MetalError` before `newBufferWithLength` (today a nil there is checked, but an over-limit-but-allocatable size never errors).
3. Track pooled bytes; when the total would exceed `recommendedMaxWorkingSetSize`, run `drop_unused_buffers()` and error if still over — converting the silent corruption into a catchable OOM so `vae_tiling::decode_with_oom_fallback` can do its job.
4. Check/propagate `ResidencySet` `commit()` failures instead of discarding them.
5. Perf aside discovered en route: `allocate_zeros` fills `buffer.length()` — the whole rounded bucket — so a 9 GiB request in a 16 GiB bucket blits 16 GiB of zeros.
6. Fix the independent `uint tid` wrap in the im2col kernel (latent ≥ 171) while in there.

## Acceptance

- Repro ladder and full-decoder repro pass whole at latent 128 F32 on Metal (see `tmp/vae-metal-repro/`; an allocator unit test pinning `buf_size` behavior belongs in candle).
- Z-Image 1024² whole decode on Metal matches CPU reference (rel < 1e-3) with real weights.
- Then, in mold: delete `MAX_WHOLE_VAE_DECODE_LATENT_AXIS` + `requires_tiled_vae_decode` and the proactive tiling call sites, keeping the reactive `decode_with_oom_fallback` ladder; re-run the #946 bisection as UAT.

## Context

- #946 (original bisection + workaround), and the Phase-1 PR that reduced the workaround's cost and gated it to Metal.
- Upstream evidence the VAE itself is fine: stable-diffusion.cpp decodes the identical latent-128 FLUX VAE whole on Metal (`docs/z_image.md` example), diffusers' FLUX `AutoencoderKL` config never tiles at this size.

jamesbrink:
Z-Image could silently return partial or zero-filled 1024-pixel VAE output on Metal. This pins the bounded Metal convolution backend from utensils/candle#12, removes the temporary latent-120 proactive tiling cap, and completes every Metal decode attempt inside a bounded OOM recovery policy before accepting pixels. CPU and CUDA retain their existing decode and fallback behavior.

The Metal recovery contract is isolated behind injected tests: recognized OOMs may retry with tiles, repeated OOM during cleanup cannot suppress the existing eager CPU fallback, and unrelated errors propagate. The Candle pin is synchronized across the workspace and desktop roots; mobile has no Candle packages in its lockfile. The H3 source marker records the newly compiled Candle identity without relabeling historical qualification evidence.

Validation:

- 11 recovery, 17 VAE-tiling, and 35 Z-Image pipeline tests pass with Metal enabled.
- Metal inference Clippy, workspace formatting, desktop/mobile locked metadata, CI routing, and single-Candle-identity checks pass.
- Candle PR #12 is green on macOS, Linux, Windows, Clippy, rustfmt, and secret scanning.
- Whole real-weight VAE outputs from 768 through 1024 pixels match CPU globally and per quadrant (relative L2 2.35e-6 to 5.24e-6, gate 1e-3).
- A real generated 1024 latent matches CPU at 3.59e-6; reference, CPU, and Metal images were individually inspected with no visible corruption.
- Independent source reviews found no Candle or Mold blocker.

The qualification does not claim 1152-pixel support: that attempt was refused by the 12 GiB native allocation ceiling. A final production-planner render was not launched because the safe source route is sequential, while the current local planner deliberately selects eager loading for Z-Image GGUF; a custom harness would not prove the shipped planner path. Details and retained evidence are in `docs/qualification/zimage-metal-vae.md`.

Closes #1040 after the dependent Candle PR is accepted and this PR passes terminal review/CI.


jamesbrink:
@greptileai review

PR #1626, opened 2026-09-07
fix: painted 3-D posters, transparent turntables, and three UI corrections

jamesbrink:
Six fixes reported from the desktop app, on one branch, one commit each.

### 1. Painted 3-D prints rendered grey (`2db9730c`)
A Hunyuan3D print with baked PBR textures showed the bare placeholder surface in its gallery tile and in every turntable frame, while the Lightbox's 3-D viewer showed it in colour.

The rasterizer had **no albedo input at all**: `poster::shade` multiplied one hard-coded near-white constant by a two-light Lambert term, and `read_glb`'s own doc comment said "Material images are not part of the returned geometry". The viewer looked different because it is a separate WebGL implementation that *does* read the embedded PNG.

`read_glb_scene` reads the geometry with the file's `baseColorTexture` and `baseColorFactor` — mirroring `readBaseColorTexture` in `studio/lib/glb.ts` so both renderers take the same image out of the same file — and `poster::Appearance` carries it into the shader, which interpolates UVs and `COLOR_0` through the `face_ids` and `barycentric` planes the G-buffer already wrote.

Bare geometry keeps the constant-albedo ramp **byte for byte**: the placeholder grey deliberately matches `MESH_PLACEHOLDER_SVG`, and mold stamps a dark `0.22` `baseColorFactor` onto unpainted geometry that must not repaint every existing poster. `MESH_POSTER_REVISION` bumps to `p3` so cached grey tiles re-render.

### 2. Transparent turntable exports (`994a518c`)
New **Background · Transparent** checkbox on the 3-D export sheet, remembered for next time; also `mold library export --transparent`, the `export_mesh` MCP tool, and `POST /api/gallery/export/:filename`.

`shade` now works in RGBA throughout — an opaque frame drops the alpha in `downsample` and is unchanged, a transparent one resolves through `downsample_rgba`, which averages premultiplied so a subpixel that missed the mesh does not drag every edge toward black. APNG and WebP keep the antialiased outline; GIF thresholds at the midpoint and **disposes every frame to the background**, without which a turning object smears its own trail through the cleared area.

Refused on a geometry transcode and on a video, and sent only when on, so an untouched export posts the body it always did.

### 3. The composer's Shape chip did nothing (`dd8489ca`)
It emitted `open-shape`, whose only handler set `inspectorTab = "settings"` — the value it is initialised to. The `▼` advertised a menu that did not exist. Its label also came from `outputFamilyLabel(width, height)` while the rail's came from `resolveOutputShape`, so the rail could highlight **Source** while the chip named a size, and a size the rail marked `≈` the chip stated flatly.

`useOutputShape` is now the one resolver both doors read *and write* through. The chip opens the same `ShapePicker` and `ResolutionSelector` the rail has, seated in a `shape` slot exactly as `style` already is.

### 4. The 3-D face budget (`34c46b20`)
"Fewer faces load faster in other apps" said why you might want it, never what it does. It now says it merges flat areas down to a triangle ceiling while creases and color survive. The input was also narrower than its own placeholder however wide the rule declared — `.ms-seed__input` gives it `min-width: 0` while the label column keeps `min-width: auto`, so it lost every flex-shrink contest. The phone carried the same control under the engine's words ("Target faces", "decimates") and now matches.

### 5. The queue progress bar didn't move (`f24ec304`)
A row captioned "Generating PBR views · 11/15" drew a bar at ~8%. `rowStatusLine` reads a shared row's `current`/`total`, but both meters understood only `row.print.step`, a field only `DenoiseStep`/`Preview` events populate — which a paint stage never emits, and which a fleet or post-reconnect row has none of anyway. `rowProgressFraction` is now the one authority, beside the sentence it has to agree with.

### 6. Pausing one job read as pausing the queue (`e3f99a22`)
**Dispatch was never at fault** — `set_owned_job_paused` moves one row by id, `claim_next` only claims `state = 'queued'`, and the per-job route never touches `QueuePause`. The wording was: a restart-parked row and a row someone paused both wear `state = 'paused'`, `explicitly_paused` never left mold-db, and `queueWaitLabel` gave both "Paused after restart" — the more alarming of the two, on the row the user had just paused.

The bit now rides `GET /api/queue`, additive and absent where a host cannot distinguish them (for which "after restart" was always true). Regression tests pin all three scopes: per-job pause/resume hit the per-job route and never the host gate, a machine pause moves no row, a global pause parks no row, and a host without the capability — or a running row — offers no entry.

---

**Verification.** `cargo clippy --workspace --all-targets` clean; `cargo test -p mold-ai-inference --lib hunyuan3d` (240), server thumbnails/queue/routes suites, `studio` (1346), `desktop` (6295), `web` (1755) all pass; `vue-tsc -b` clean on desktop and web; prettier, knip, frontend-architecture, the prompting-guide and generation-profile contracts, the routing contract, and `website/ bun run verify` all pass.

Not covered by automated tests: the actual pixels. A painted print's poster and a transparent GIF want a look on real hardware.

https://claude.ai/code/session_01AJFJBfYJTC39zcTvGHYrFE


jamesbrink:
UAT on the painted 3-D path, run against a real print (`mold-hunyuan3d-2.1-fp16-1788744717014.glb`, a 2048-square painted Hunyuan3D 2.1 mesh) driving `glb` → `poster`/`turntable` → GIF with this branch's code.

**PBR — works.** The reader finds the material (`baseColorTexture` 2048², `baseColorFactor [1,1,1]`, `uvs=true`) and the sweep renders in the object's own colours: mean opaque RGB `[70, 33, 90]`. The grey placeholder would be r≈g≈b.

**Gallery — correct.** Row present as `format=glb`, `model=hunyuan3d-2.1:fp16`, seed recorded, not trashed; metadata carries `mesh.texture: true`, `texture_resolution: 2048`, and `prompt: ""` as a prompt-`Ignored` family should.

**Transparency — works.** Decoded GIF frame 0 has `clear=51454` against the renderer's `clear=51454`, an exact match, and zero holes in the object.

**One claim in this PR was wrong and is corrected by ca99d066.** c5ed9322 said the transparent palette entry pulled a dark painted mesh's texels onto itself and punched holes in it, and stamped cleared pixels magenta to stop it. That mechanism does not exist: `color_quant`'s `search_netindex` adds the alpha residual to the squared distance at equal weight to R/G/B, so an opaque pixel is 255² from a transparent entry and always resolves to an opaque one first. A/B on the real mesh gives `HOLES=0` and `clear=51454` with either clear colour.

The stamp is kept, for the reason that does hold: `Frame::from_rgba_speed` takes its transparent colour from the *last* sub-cutoff pixel in scan order, and only fully cleared pixels arrive as `[0,0,0,0]` — the antialiased edge carries object RGB under low alpha — so without a stamp the transparent entry is render-dependent. Its colour is now black rather than magenta, because a decoder that ignores transparency composites it.

Not covered: a fresh generation on this branch (no commit here touches the shape or paint pipeline, only read/render/encode), and the desktop app's own Library tile and export sheet, which need a human's eyes.

Note for future UAT: plato cannot host this one — it reports `mesh.textures: false` and serves an older `git_sha`, and `transparent` is an additive field an older host ignores rather than refuses.

PR #1657, opened 2026-09-08
feat(studio): give every theme a light and a dark tone

jamesbrink:
Pick **Nebula**, turn on **Match system**, and in daylight the app repaints as *Porcelain* — a different typeface, different corners, different accent. The theme you chose is gone. Same for Safelight. That is what `THEME_PAIR` did:

```ts
safelight: { dark: "safelight", light: "porcelain" },
nebula:    { dark: "nebula",    light: "porcelain" },
```

Tone was baked into theme *identity*, so each of the six themes was locked to one, and the partner had to be picked by hand.

## The change

A theme is its typography, radii, density and accent hue. Light-vs-dark is a tone that identity is rendered in. So a `ThemeId` is now `` `${family}-${tone}` `` over five families, tone is **derived** from the suffix, and a theme's partner is that same theme in its other tone.

**5 families × 2 tones = 10 palettes, replacing 6.** Six of the ten maps are today's palettes moved, not redrawn. Four are new:

| Family | Dark | Light |
| --- | --- | --- |
| **Mocha** | unchanged | **new** — Catppuccin *Latte*, the flavour's own light half |
| **Safelight** | unchanged | **new** — the darkroom with the lights on: warm paper, amber press |
| **Blueprint** | **new** — cyanotype: prussian ground, white drafting lines | unchanged |
| **Graphite** | unchanged | today's **Porcelain**, re-set in the family's IBM Plex |
| **Nebula** | unchanged | **new** — bone paper, oxblood ink, crimson press |

Graphite and Porcelain were already one theme under two names — same greys, same IBM Plex Mono, same radii, same density, differing only in tone. They merge. **Porcelain retires as a name, not as a palette.**

Every new palette was verified against the existing rules in `tokens.contrast.test.ts` — not by eye. Latte needed its ink darkened one step and its control edge mixed at 60% (its ink is lighter than the other light themes', so the standard 52% misses 3:1).

## The picker: two independent controls

Entries name **only the theme** — no label anywhere carries a tone, which retires `toneLabel` and the web select's `Mocha · dark` — beside one **System · Light · Dark** control. Both project onto the same persisted `{ theme, matchSystem }`, so choosing a theme never moves the tone and choosing a tone never moves the theme. The standalone Match-system switch folds into System.

## All four surfaces

Desktop, web and the phone already shared `ui/tokens.css`. **The TUI did not** — it ran eleven unrelated presets with no code, test or generator tying it to the design system, so its `Mocha` was Catppuccin's rather than the app's and `StudioDark`'s accent was a magenta no GUI theme has ever shown. It now carries the same ten, derived from the shared tokens and pinned to them by a test. Every retired slug (`studio-dark`, `ristretto`, `gruvbox`, `tokyo`, `nord`, `dracula`, `latte`, `porcelain`) still loads and maps to the nearest family.

Its Appearance panel gains a Light-or-dark row but **no System**: a terminal has no OS appearance to follow — the emulator owns everything outside the frame — and offering a choice that cannot resolve would be a lie.

## What this deletes

The pairing table was maintained by hand in four places (`ui/theme.ts`, the Rust `light_partner()`, and the two inline pre-paint scripts). All four collapse to swapping a suffix.

It also adds the guard that never existed: `web/index.html` and `desktop/index.mobile.html` are hand-written second implementations of `resolveTheme`, and nothing checked them. `web/src/lib/prePaintTheme.test.ts` now runs the real script text against the real contract over **82 cases** — every saved shape either surface can hold, old and new.

## Migration

One table, chained after the existing `{ family, theme }` step so a pre-redesign install lands in one hop:

`mocha` → `mocha-dark` · `safelight` → `safelight-dark` · `blueprint` → `blueprint-light` · `graphite` → `graphite-dark` · **`porcelain` → `graphite-light`** · `nebula` → `nebula-dark`

The Rust side rewrites the raw JSON *value* before the typed parse, so an old `"porcelain"` can never fail `AppSettings` and drop saved machines (the 2026-09-03 incident).

## Verification

- `tokens.contrast.test.ts` — 62 assertions across all ten maps, unchanged rules
- desktop 6484 tests · web 1927 · TUI 837 · Rust settings 39 — all passing
- `check:frontend`, `check:architecture`, `check:dead-code`, clippy `-D warnings`, prettier, `cargo fmt`
- Real-terminal UAT: five cards, the tone row, `theme-set nebula-light` / `safelight-light`, and the choice surviving a relaunch

## Open for review

**Graphite's light accent.** Its dark tone signals in amber, but amber can't carry white text. This PR keeps Porcelain's deep teal (`#0f766e`) — zero existing pixels change. Carrying the family's hue across instead means a bronze (`#a3540a`, with warning/star moved to `#7a5b00`); both pass every rule. The interactive mockup toggles between them.

Also note the merge moves the light tone's sans from Manrope to IBM Plex Sans and onto Graphite's slightly roomier scale — a family has one typeface. Colours are unchanged.

https://claude.ai/code/session_01S3zNyVnTJmXT45QnG2BcvJ


jamesbrink:
## Peer review (codex, `--base main`) — two findings, both real, both fixed in 996adff

**P2 — `matchSystem` lost when renaming a theme.** `migrateLegacyTheme` answered `matchSystem: false` for a renamed id, but the pre-tone shape already carried that flag *beside* the id. Every web and phone install with Match system on would have lost it on upgrade — and lost it invisibly: the inline pre-paint script reads the flag directly and painted correctly, then mounting flipped the palette back.

The function now takes the saved flag, so one place owns the whole decision. Notably the workaround in `prePaintTheme.test.ts` — commented *"migrateLegacyTheme cannot see a sibling flag; the script can"* — was the smell that predicted this exact bug, and it's now gone: the guard compares against the plain function with no compensation.

The desktop was never affected. Its Rust migration returns `None` for the flag rather than a value, which leaves the saved one alone — the right shape, and the reason the two implementations diverged.

New coverage: the shared contract, plus the failing shape driven through both real loaders (`web/src/lib/theme.ts`, `loadMobileSettings`).

**P2 — the UAT tone-row focus walked the card grid.** It reached the Light/Dark row by pressing Down from the cards, which live-applies each card it passes. At the usual four columns, `theme-set mocha-light` from a dark theme selected Mocha, walked to Nebula, then toggled the tone — landing on `nebula-light` and reporting failure. It now approaches from below (Configuration, then up), which touches no selection, and `theme-set` does the tone first so the family cycle is the last thing that runs.

Verified in a real terminal, all six transitions in both directions:

```
theme-set mocha-light      PASS: settings.tui.theme = mocha-light
theme-set graphite-light   PASS: settings.tui.theme = graphite-light
theme-set nebula-dark      PASS: settings.tui.theme = nebula-dark
theme-set safelight-light  PASS: settings.tui.theme = safelight-light
theme-set blueprint-dark   PASS: settings.tui.theme = blueprint-dark
theme-set mocha-dark       PASS: settings.tui.theme = mocha-dark
```

Gates after the fixes: desktop 6486 · web 1928 · TUI 837, typecheck, knip, architecture, prettier, `cargo fmt`, clippy `-D warnings`.

https://claude.ai/code/session_01S3zNyVnTJmXT45QnG2BcvJ

PR #1682, opened 2026-09-10
ci(docs): install ripgrep so the env-var coverage check actually runs

jamesbrink:
The website verifier harvests `MOLD_*` env-var names with ripgrep and checks each is documented under `website/`. On CI that check has been silently dead: the runner has no ripgrep, and the harvest was a shell pipeline whose exit status was `sort`'s, so a missing `rg` still yielded an empty set and the check passed vacuously with no warning. That is how `main` stayed green with `MOLD_DESTINATION_API_KEY` undocumented until a local run caught it in #1679.

- `docs` job in `ci.yml` and the Pages workflow install ripgrep before `bun run verify`.
- `verify-docs.mjs` runs `rg` alone (filtering in JS) so a missing binary throws.
- Under `CI` a missing ripgrep is a failure; locally it stays a warning.

Verified locally: normal run passes (259 env vars harvested, all documented); with a failing `rg` shim under `CI=1` the verifier fails with the new message; without `CI` it warns and passes. `actionlint` clean.

https://claude.ai/code/session_016L3TinDUYUKsUAWVAFi9kQ

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -318,6 +318,8 @@ jobs:
       - name: Install docs dependencies
         working-directory: website
         run: bun install
+      - name: Install ripgrep for the env-var coverage check
+        run: sudo apt-get update && sudo apt-get install -y ripgrep
       - name: Check docs formatting
         working-directory: website
         run: bun run fmt:check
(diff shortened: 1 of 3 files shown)


jamesbrink:
Closing: folding this into the Rust CI speed-up PR rather than spending a separate full run on it.

PR #1689, opened 2026-09-10
ci(rust): two parallel Rust lanes on main, MSRV weekly, sccache out

jamesbrink:
## Why

A push to `main` ran one 39-minute `rust` job, sequentially: MSRV (4.2 min), default-feature clippy (2.6), `cargo test --workspace` (15), the H3 arms, the private bins, and the union clippy and nextest builds. Nightlies never waited on it (release.yml builds from every main push and does not gate on CI), but it held a runner slot for the whole chain and was the slowest signal on `main`.

## What changes on `main`

- **Two parallel lanes.** `rust` keeps the H3 arms, the private bins, and the union clippy + nextest builds (the ones that warm the cache PRs restore). A new push-only `rust-default` job runs the default-feature clippy and the complete suite through nextest's new `main` profile (every test, live catalog monitor included, one retry that reports FLAKY instead of red). Expected wall clock: about 15 minutes per lane instead of 39, with fewer total runner-minutes than before.
- **One cache writer.** `rust-default` restores `workspace-default` with `save-if: false`. Two writers would race, and a PR could restore a cache with no union artifacts. Its `env` mirrors `rust` so rust-cache computes the same key.
- **MSRV moves to `.github/workflows/msrv.yml`:** weekly (Mondays 13:17 UTC), `workflow_dispatch`, and on any manifest or lockfile change to `main`. No rust-cache, because the repository's cache store is at 9.58 of 10 GB and a new key would evict another job's cache.
- **sccache leaves the Rust job.** The store holds zero sccache objects, so the probe failed on every run inspected, and the probe's on/off toggle of `RUSTC_WRAPPER` changed the rust-cache key (it hashes every `CARGO*`/`RUST*` value), so a PR could miss the cache `main` had just saved.

## What does not change

- PR behaviour of the required `rust` check: same diff-scoped selection, same `pr` nextest profile.
- Nightly/dev delivery: `release.yml`, `desktop.yml`, `windows-nightly.yml`, iOS and Android all trigger on every push to `main` independently of CI.
- Coverage and Windows typecheck jobs are untouched (coverage still carries sccache under its own key).

## Contract and docs

`scripts/tests/ci-routing-contract.sh` pins the `rust-default` lane (push-only, same shared key, `save-if: false`, default clippy, `cargo nextest run --profile main --workspace`), the `main` profile with no `default-filter`, the MSRV workflow (schedule, dispatch, lockfile path, both `cargo check` invocations), and the absence of `sccache-action` from both lanes. `scripts/ci-local.sh` runs the same nextest command (falls back to `cargo test` without cargo-nextest). `CLAUDE.md` and `.claude/rules/release-ci.md` describe the new shape.

## Verification

- `bash scripts/tests/ci-routing-contract.sh`, `bash scripts/tests/ci-affected-packages.sh`, `actionlint` on every workflow: pass.
- `cargo nextest run --profile main --workspace` run locally in full (result noted in the PR conversation).

## Follow-ups not in this PR

- nextest never runs doctests. The workspace has none today; a future doctest would be silently skipped unless a `cargo test --doc` step is added.
- Main's first run after merge is cold for the `rust` job because dropping `RUSTC_WRAPPER` changes the cache key once.

https://claude.ai/code/session_016L3TinDUYUKsUAWVAFi9kQ


jamesbrink:
Local verification of the main lane's command, `cargo nextest run --profile main --workspace`:

- First run: 9,816 tests, 9,815 passed, 1 failed on both tries: `gpu_worker::tests::drain_after_model_ready_finishes_generation_then_stops` at "generation must start only after model readiness: Timeout". It passes alone. Eight worker tests waited one second for a barrier the thread had already been handed; under nextest's process-per-test scheduling that is too tight. Commit bdcf49fa replaces the eight literals with one 30-second `BARRIER_DEADLINE`. Ordering assertions are unchanged.
- After the fix, `mold-ai-server` under the same profile: 2,248 passed, 10 skipped, no retries used.

The `rust-default` job is push-only, so its first real run is main's run after merge.

PR #1723, opened 2026-09-16
fix(gallery): prevent desktop abort on historical trash re-import

jamesbrink:
A background gallery re-import could abort the entire desktop app when an older trash operation had moved a print's bytes and database row but left its committed archive live. Purging the trash did not retire that authority; publication then failed with `multiple live committed archives claim gallery filename`, and every restart hit the same unresolved commit.

Retire verified historical trash authority before deleting its evidence. Preflight conflicting imports under the namespace lock before entering the fatal commit phase, and clean up unpublished refusals. Recover an interrupted re-import only when its checksum, size, format and generation descriptor match the existing archive, allowing a previously absent elapsed generation time to be filled in. Preserve the original archive identity and retained source-media bindings; changed inputs or conflicting files remain refused.

Validation:
- Gallery-import route tests, including historical archived trash with a changed replacement descriptor.
- Transaction regression tests for interrupted recovery, retained-media identity, pre-commit conflicts and preservation of unverified files.
- Replayed a private, relocated copy of the actual incident's archive/journal and video: one attempt rolled forward, the next restart required no recovery, and the video checksum was unchanged. The live gallery was not modified.
- Server Clippy with warnings denied and Rust formatting.


jamesbrink:
@greptileai review

PR #1728, opened 2026-09-17
feat(macos): native macOS app -- Library, Machines, Create, Models, Queue, Settings

jamesbrink:
## What this is

A native macOS app for mold, built on `feat/macos-native-app` since
2026-09-15 (406 commits over `main`, 1,021 files, ~98k lines under
`apps/macos`). Milestones M1–M8 built it; the **review pass of 2026-09-17/18**
(next section) audited it with six reviewers, fixed everything they found,
closed the parity gaps against the Tauri desktop app, and had the result
reviewed three more times. The branch is merged up with `main` (v0.30.1) as
of `922a8bae`.

## What it does

- **Generate** — every control comes from the chosen model's own generation
  profile, so a model added to mold tomorrow gets correct controls with no
  app change. The capsule holds the machine (Auto follows the default
  machine, else the first that's up; any machine that is up and generates
  can be pinned, and the model follows it), the shape as aspect + size from
  the recipe's own ladder, steps, guidance, seed and batch, with the estimate
  and Generate at the trailing edge of a last row of its own, where it stays whatever appears beside it. An Expand button under the
  prompt rewrites it or suggests other ways to say it, in place, with the
  original kept. The source well and the reference strip take a file, a
  print dragged out of the Library, or one chosen from a Library sheet.
  Durable submission, live step progress, a denoise preview, then Save /
  Copy / Show in Library. Generate never turns into Stop: a second press
  queues another batch and the pane follows them in turn.
- **Library** — every machine's prints in one day-sectioned, host-badged
  timeline: search with real tokens, sort, collections merged across the
  fleet, trash with a countdown and Put Back, and favourite/tag/file/rename
  all undoable from the Edit menu.
- **Queue** — every machine's work live from its event stream: drag or
  Move Up/Down to reorder, Empty Queue, a held row that explains itself in
  words and offers Retry or Try Again, Move to… another machine, and Cancel
  Job on every row including held ones (the trailing ✕, the contextual menu,
  or ⌘⌫), a Dock badge and notifications.
- **Models** — an Installed table grouped by family with the manifest's own
  plain-English trade-off, Discover through the machine's own catalog,
  install/repair/load/unload/delete, and a gated model's licence accepted in
  place.
- **Machines** — every machine's page: its GPUs and what each is holding, a
  lifecycle switch where the scheduler will honour one, live memory and CPU,
  what's queued and installed there, and — on a keyed machine this app holds
  an operator key for — its paired phones, with Pair a Phone… and Revoke.
- **Settings** — nine tabs: curated panes for what people actually change
  (General — with Appearance: System, Light or Dark, applied to every
  window — Generation, Expansion, Library, Performance, Accounts, Machines,
  This Mac), plus one honest Advanced table that renders every row the
  machine's config API returns, from that row's own JSON type.

## The review pass (2026-09-17/18)

Six Opus reviewers audited the M1–M8 app against the server, the Tauri
desktop app and the HIG: **15 HIGH, ~45 MED, ~25 LOW** plus a parity matrix.
The reports, the plan, its recorded deviations and one ledger per lane are in
the repo at `apps/macos/docs/review-2026-09-17/` (`01`–`07`, `PLAN.md`,
`STATUS-*.md`). Every row is closed with a commit and a test; nothing was
deferred.

| Lane | What landed |
| --- | --- |
| A · wire | reorder PATCH in the server's own index space; `.unknown` open enums never re-encoded; query escaping and `escaped()` on every interpolated id; stable merge sort; SSE refusals keep the server's sentence and licence payload; `X-Api-Key` stripped on cross-origin redirect (`RedirectGuard`); media tickets on keyed hosts; one stream buffering policy; response size ceilings |
| B · generate | layout from `source_relation` (replaces / exclusive / combines); result canvas per kind (clip, picture, mesh); Stop during submit; length ceiling and `text_only_auto_chain_refusal` wording; `fit()` honours `off_bucket` and rounds down; identity photos transcoded; file I/O off the main actor |
| C · library | one `SafeFilename` at decode and every write site; events queued during an outbox; day grouping follows Sort By; modifiers from the key event; undo grouped by target; cache budget never evicts a file in use |
| D · queue/models/machines | pairing `expires_at` in seconds; every reconnect emits `resyncRequired`, wake forces one, a poll where the host wants one; `can_pause_job` / `cooperative_cancellation` read; single-flight hydrate |
| E · secrets/settings | Keychain retired for a 0600 `secrets.json` (a port of the desktop's `secrets.rs`, one-time migration); explicit set/clear per host; plain confirm dialog on machine removal; `#if DEBUG` around every UAT hook |
| F · engine/release | the loopback engine is **keyed** (minted `local-engine-api-key`, non-web CORS origin); FFI `catch_unwind` + drop guard; `.running` only after `/api/status` answers; quit honours the server's 45 s budget behind a Finishing… sheet; linkage fix-up before signing; `release` depends on `engine`; `macos-native.yml` CI |
| F1 · mesh | native Metal GLB viewer — a line-for-line port of `MeshViewer.vue`'s GLSL and `meshViewerCamera.ts` (a Rust contract test reads the Swift file); orbit/zoom/keys/auto-rotate; exports from `capabilities.mesh.export_formats`, geometry options from `export_geometry`, turntable sheet |
| F2 · reuse | `OutputMetadata` widened to everything `applyMetadataToForm` reads; every copy probed for retained source media; same-host reuse session, cross-host relay; the print's picture placed in the well; the seed locked on reuse (web parity) |
| F3 · upscale/activity | Make Bigger… on stills and clips (`video_upscale`), 750 ms poll behind a fence; `ActivityStore` with `activity.ts`'s reconcile rules; "Also running" rows with stop notes where the server says `can_cancel: false`, counted on the machine card |
| F4 · controls | IP-Adapter weight, scheduler, CFG++ (the one sanctioned family set, pinned against `generationCapabilities.ts`), Wan recipe, LTX-2 guidance overrides; `source_fit`; whole-queue pause; long clips as **ephemeral chains**; versioned draft persistence |
| F5 · updater | Sparkle 2, feeds on the existing GitHub release channels, Settings ▸ General channel picker, `make appcast`, `macos-native-distribution.yml`; **the EdDSA key is installed** (public half in `project.yml`, private half in `MOLD_NATIVE_SPARKLE_KEY`) |
| G–K | cleanup; sidebar; Machines fleet overview; inspector; every "plato" replaced by `workstation` with a source-scan test |
| L | type budgets — `GenerateController` 603→446, `LibraryActions` 657→566, `LibraryStore` 779→541, `RenderDraft` 915→500; a heartbeat lost-wakeup fix |
| M | **one picture chooser** (`PictureWell`) behind the source well, the reference strip, ControlNet and identity — file, Library print, paste or drop through the same menu on every one; captioned wells |
| N | the native `.inspector` (the divider carries up through the toolbar, the search field ends at it); section content aligned to its title; **one error voice** — what did not happen, the machine's reason, and the way forward (`failureSentence`), pinned by a table test and a source scan |
| UAT | 50 items driven on the real app against hal9000 and a keyed host: 40 PASS, 10 FAIL, all ten fixed (AVKit link crash on a clip, Quick Look size, queue menu on a batch of one, Share placement, typed search tokens, pending leftovers on machine removal, …) plus every screenshot the owner sent that evening |

Three final reviews on the finished range (two Opus, one Codex astra) each
found one real defect, all fixed: the inspector toggle's reserved width was
also its hit region; the engine's own probe and shutdown requests carried the
key with no redirect guard; a reused print's picture download could land over
a prompt edited meanwhile.

## Milestones (M1–M8, before the review pass)

| # | Milestone | Commits | Closed |
| --- | --- | --- | --- |
| M1 | Library depth | 8 | 2026-09-16 |
| M1.5 | Structure (the post-M1 audit fix plan) | 12 | 2026-09-16 |
| M2 | Machines | 6 | 2026-09-16 |
| M3 | Create I (expansion, File under, output, upscale) | 9 | 2026-09-17 |
| M4 | Create II (LoRA, identity, inpaint mask, ControlNet, clip) | 10 | 2026-09-17 |
| M5 | Models (Installed table, Discover, downloads, licences) | 10 | 2026-09-17 |
| M6 | Queue (reorder, transfer, batch grouping, badge/notify) | 12 | 2026-09-17 |
| M7 | Settings + polish (curated panes, Advanced, pairing, menu bar, a11y) | 10 | 2026-09-17 |
| M8 | The Generate capsule, redesigned after the user reviewed it (machine + shape in the row, the actions row, visible Expand, Library picker, queued Generate, popover fix), plus Settings ▸ General ▸ Appearance and Cancel on held queue rows, and a held batch no longer followed or re-attached at launch | 8 | 2026-09-17 |

Commit counts are the number of commits recorded against each milestone's own
ledger table in the plan (`i-have-a-totally-shiny-tarjan.md`); a row logged
only as a live capture with no commit hash is not counted. Dates are the
plan's own stated close dates for each milestone. M8 was asked for after this
PR opened, from a screenshot of the capsule, and pushed onto it.

## The decisions that shaped it

1. The protocol seam was **widened**, not deleted — `MoldBackend` grew to
   cover every route the UI actually calls, rather than the app admitting it
   has one transport.
2. `UndoManager` registration lives in the **store**, never at call sites —
   the store is the one thing still holding the previous value.
3. **One `/api/events` connection per machine**, fanned out to every store,
   replacing per-feature polling.
4. **SwiftUI's own `.inspector`** (since Lane N) — it is the only thing that
   carries the divider up through the toolbar; the column is exactly as wide
   as the search field macOS pins to the toolbar's trailing end, so nothing
   straddles the divider. Per-pane `@AppStorage` keeps two independent states.
5. **A batch is N one-output requests**, never `batch_size: N` — the server
   refuses any child whose `batch_size` isn't 1.
6. **Identity conditioning is positive-only and parked, never disabled** — an
   unqualified checkpoint renders nothing extra and refuses nothing; a staged
   photo waits for a qualified model.
7. **The queue's reorder index is not the listing's `position`** — the PATCH
   indexes the machine's `queued` rows alone, and a batch moves as ascending
   single-row calls.
8. **A transfer's admission body is assembled from bytes** — the export is
   spliced in opaque, because decoding and re-encoding it through this app's
   request type would drop the media it carries.
9. **Advanced renders from the value's JSON type** — the wire declares no
   schema at all, so the curated panes carry their own knowledge instead,
   pinned to the Rust config registry by a test.
10. **Pairing is issuer-only** — the app can pair a phone to a keyed machine
    it already holds an operator key for, but it is never itself a claimant,
    and This Mac's own engine can never be paired (keyless on loopback).
11. **Generate never turns into Stop** — the host's queue is durable and one
    press is one batch, so a second press while a render runs admits another
    batch instead of being refused; the pane keeps its own FIFO and follows
    them in turn, each on ITS machine's connection (the Machine control can
    move between two presses). Stop stops the one on screen; Stop All Queued
    is in its menu.
11a. **The actions row is its own row, always.** The first cut pinned the
    hint and Generate to the last wrapped control row; the layout's width
    was the widest row, so Generate stopped short of the capsule's edge, and
    Stop appearing could push it down a row mid-run. Now the hint truncates
    at the leading edge and Generate holds the trailing one, a sheet's
    button row, and nothing that appears beside it moves it.
12. **Two remembered machines with two meanings** — `defaultMachine` (where
    work goes, what Auto follows) and `selectedMachine` (what you are looking
    at); the capsule's Machine control pins a third, per-draft choice on top,
    and a Reuse pins the print's machine because the model was adopted there.
13. **The Shape ladder is the server's** — aspects come from the recipe's
    `aspect_groups`, labelled the way `generation_profile.rs` labels them;
    an off-ladder size carried in by a Reuse is shown as its own row, never
    snapped.

## The incidents, in full and unsoftened

- **M3, 2026-09-16: two probe POSTs to hal9000.** The design agent sent two
  test requests to `/api/generation-batches` on hal9000, a live host. One was
  refused at 422 (`batch_size: 2`). The other was **admitted at 202** with a
  nonexistent model and cancelled immediately — nothing rendered and no
  gallery row exists — but it left one prompt-history row,
  `{"prompt":"probe","model":"not-a-real-model-probe"}`, on hal9000. That row
  cannot be removed without clearing hal9000's entire history, and it has not
  been cleared. This was reported to the user at the time; every later brief
  in this plan says "no POSTs to any live host" in those exact words.
- **M4 S1: a test-run crash.** The first app-bundle test run for M4 S1 stopped
  after five tests. `FakeBackend.callCount` was being read while a store task
  appended to it concurrently — a real data race, not a flaky test. Fixed by
  recording the count under an `NSLock`.
- **Two accidental `git stash` uses.** The M3 S2 agent (building
  `Expansion`/`ExpansionOffer`) used `git stash` to get a clean red-test
  baseline despite being told not to; it was popped immediately and no stash
  was left. The M5 S2 agent (building `ModelStore+Manage`/
  `DownloadStore+Install`) did the same and disclosed it unprompted; also
  popped at once, tree whole. Both are recorded in the plan's own ledger.

A PR that hides incidents like these is worth less than one that lists them.

## Manual checks a human still owes

- **The mask sheet's `UndoManager`, with a real mouse.** Every automated
  attempt to drive the sheet's `DragGesture` by posted mouse events failed to
  register a stroke, so which `UndoManager` the sheet resolves against is
  unverified by anything but code reading. Ten seconds is enough: open the
  mask editor, paint one dab, open Edit — it should read "Undo Stroke".
  (`⌘Z` itself is proven working, through the sheet's own hidden undo button;
  it's specifically the Edit-menu wording that needs a real mouse.)
- **Two things only a real mouse can do on the new capsule.** Click one row
  of the expansion popover and see it highlight (the fix is a
  `simultaneousGesture` for the double-tap that used to eat the click; AX
  cannot click a List row, so it is verified by reading), and drag a print
  from the Library onto the source well (a drop was never driven by PID).
- **Right-click a held queue row.** Its new contextual menu (the retry
  choice, Move to, Cancel Job) is pinned by a pure `menuTitles` test, but
  `AXShowMenu` on a SwiftUI `List` row is unsupported, so no capture shows it
  open; the row's ✕ was photographed from the fixture.
- **VoiceOver over the four sites S7 labelled**: the dismiss button in
  `Shell/FailureBanner.swift`, the remove-tag button in
  `Generate/FileUnderGroup+Tags.swift`, and the photo wells in
  `Generate/IdentityGroup.swift` and `Generate/ReferenceStrip.swift`. All four
  now carry a label or `.help`; nobody has run VoiceOver over them since.

## Honest advisories

- **`make lint` prints one `large type:` line**, `HTTPBackend` (1,402 lines
  across its files). It stands because Swift has no conformance delegation:
  splitting it into composed sub-types would cost forwarder lines that the
  same lint rule would sum straight back onto the total. `LibraryStore` and
  `GenerateController` were brought under budget by Lane L.
- **`lint-a11y`'s blind spot.** It's a per-FILE floor — a file with a
  `systemName` image must also contain a label, `.help(`, or an opt-out
  comment — not a per-site audit. It cannot see whether a modifier lands on
  the glyph that needs it rather than a sibling (a tooltip on a row does not
  name the button inside it), so a file can pass the rule and still be wrong.
- **The placement hint shows the server's reason verbatim.** plato answers
  `insufficientvram` as the placement `reason` for FLUX.1 Schnell Q8 at
  1024², so the capsule reads "⚠ insufficientvram" — a code, not a sentence,
  and the app shows what it is told rather than guessing at a translation.
- **The Result strip's thumbnails fetch full media**, not a rendition route.
  This was a deliberate M3 S6 call, not an oversight; a dedicated rendition
  endpoint for strip thumbnails is a real gap but was scoped out as polish.
- **The embedded engine is keyed but still loopback-only.** Lane F mints a
  `local-engine-api-key` and exports a non-web CORS origin, so no browser
  page on the Mac can read it; it still binds `127.0.0.1`, so it cannot be
  paired from a phone.
- **A `DELETE /api/config/:key` answers `source: "default"`, but the listing
  reports a key's storage SURFACE**, so a re-read after a reset says `db`
  again. A listing cannot tell a stored value from a default
  (`routes_config.rs:48-53`); the app shows exactly what the server tells it.
- **The Edit ▸ Select All item the grid cannot enable.** SwiftUI never routes
  the system's own Select All command to a custom grid on this SDK, in two
  separate attempts (on the pane and on the grid itself). The item stays
  present and disabled, the same way Refresh does when there's nothing to
  refresh, and the grid answers `⌘A` itself instead.

## How it was verified

Every slice was gated by the orchestrator, never by the implementing agent's
own report: `make lint` (including `lint-a11y`), the `MoldClient` package's
tests (414 after M8, **920** at the tip), the app bundle's `MoldTests` (380
after M8, **758 in 108 suites** at the tip), and `make build`, all green
before a slice was accepted. Review-pass lanes ran as Opus agents in their
own worktrees with disjoint file ownership, each diff adversarially reviewed
by a second Opus agent before it was cherry-picked. Every live capture against plato or hal9000 was
**GET-only** — status, capabilities, config listings, pairing clients, model
and queue reads — never a mutation, with two read-shaped POSTs the pane has
always made: the placement preview, and `/api/expand` on plato while idle
(three calls in M3, two in M8 to photograph the fixed popover). Every write throughout M1–M7 went through
`FakeBackend` or a fixture file, with exactly one exception: the M7 carve-out,
a single config write-then-reset-then-read cycle against **This Mac's own
embedded engine**, run against a throwaway `MOLD_HOME` created for that check
alone. No image or video generation was run on any host, local or remote, at
any point in this branch's history.

The branch was rebased onto `main` (v0.30.0) once, immediately before this
PR was first opened, so the hashes the M1–M7 ledgers record predate that
rebase; `git log --grep` by subject finds each one. Since then history is
append-only: `main` (v0.30.1) was merged in at `922a8bae`, never rebased.

## How to run it

From inside `nix develop`:

```bash
macos-dev      # build and run against your real settings, logs on the terminal
macos-uat      # the same build against a throwaway prefs domain and mold home
macos-test     # package tests
macos-lint     # architecture lints
```

UAT hooks (`README.md` lists them all, every one behind `#if DEBUG`):
`MOLD_NATIVE_FRESH`, `MOLD_NATIVE_HOSTS`, `MOLD_NATIVE_DESTINATION`,
`MOLD_NATIVE_SETTINGS_TAB`, `MOLD_NATIVE_QUEUE_FIXTURE`,
`MOLD_NATIVE_PAIRING_FIXTURE`, `MOLD_NATIVE_SOURCE_IMAGE`,
`MOLD_NATIVE_LIBRARY_PICKER`, `MOLD_NATIVE_UAT_SCRIPT`.

## What a human still owes

- A first run of `macos-native-distribution.yml` proves the signed appcast
  end to end now that the key exists.
- The "plato" sweep outside the app (~140 files on `main`) is its own
  mechanical PR, never this branch.

## What is deliberately not here

- **The 3-D studio.** Out from the plan's first line — not ported, not
  stubbed. This app is scoped to generation and the library (a finished GLB
  is *viewed* and *exported* natively; it is not authored here).
- **Chain jobs.** Scripted sequences are CLI-and-API-only by mold's own
  design; there is no timeline or scene strip on any current surface.
- **Cloud provisioning (RunPod / Lambda).** A rented pod is still reachable
  as an ordinary machine by address once it exists, so nothing becomes
  unreachable — provisioning itself stays in the CLI and the web app.
- **Discord.** A different surface entirely, with its own bot and its own
  HTTP-only client crate; never part of this plan's feature gap.
- **Large reference uploads.** mold's upload-session protocol exists for
  MiniMax H3 and 3-D meshes, neither of which this app produces, so reference
  pictures always travel inline instead.
- **The claimant half of pairing.** The app already has manual host-and-key
  entry, and mobile pairing exists specifically to spare a *phone* from
  typing a key — there's no reason for this app to scan a QR code of its own.


jamesbrink:
Closing temporarily so every push in the fix pass does not run CI. Will reopen when the pass is done.

